PR #3, opened 2022-04-08
WIP: Feature/bd28

patrickskowronekdkfz:
This PR brings a new script that adds a basic auth user to the environment. Changed traefik to use environment variable instead of file

### DIFF
--- a/lib/add_bc_user.sh
+++ b/lib/add_bc_user.sh
@@ -1,15 +1,29 @@
 #!/bin/bash -e
 
 echo "This script add's a user with password to the bridghead"
+
+if [ $# -eq 0 ]; then
+    echo "No arguments provided, please provide the prject name"
+    exit 1
+fi
+
+if [ ! -f /etc/systemd/system/bridgehead@$1.service.d/override.conf ]; then
+    echo "Please create a Service first, with setup-bridgehead-units.sh"
+    exit
+fi
+
 read -p 'Username: ' bc_user
 read -sp 'Password: ' bc_password
 
-echo 
+echo
 
-bc=$(docker run --rm -ti xmartlabs/htpasswd $bc_user $bc_password)
 
-if [ -z $bc_auth_users ]; then
-    printf "Please run: export bc_auth_users=\"%q\"" $bc
-else
-    echo "Please run: export bc_auth_users=\"${bc},$bc_auth_users\""
+bc=`docker run --rm -it httpd:latest htpasswd -nb $bc_user $bc_password`
+
+if grep -q -E "Environment=bc_auth_users=" /etc/systemd/system/bridgehead@$1.service.d/override.conf ; then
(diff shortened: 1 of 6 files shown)


lib/add_bc_user.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "This script add's a user with password to the bridghead"
+read -p 'Username: ' bc_user
+read -sp 'Password: ' bc_password
+
+echo 
+
+bc=$(docker run --rm -ti xmartlabs/htpasswd $bc_user $bc_password)
+
+if [ -z $bc_auth_users ]; then
+    printf "Please run: export bc_auth_users=\"%q\"" $bc
```
torbrenner, line 12: Please move this to /etc/systemd/system/bridgehead@\.service.d/xxx.conf
Otherwise the password and user would be lost after system restart.

lib/add_bc_user.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+
+echo "This script add's a user with password to the bridghead"
+read -p 'Username: ' bc_user
+read -sp 'Password: ' bc_password
+
+echo 
+
+bc=$(docker run --rm -ti xmartlabs/htpasswd $bc_user $bc_password)
```
torbrenner, line 9: Please use https://hub.docker.com/_/httpd?tab=tags instead of xmartlabs/htpasswd

torbrenner (changes requested):

lib/add_bc_user.sh
patrickskowronekdkfz: Fixed

lib/add_bc_user.sh
```diff
@@ -1,15 +1,29 @@
 #!/bin/bash -e
 
 echo "This script add's a user with password to the bridghead"
+
+if [ $# -eq 0 ]; then
+    echo "No arguments provided, please provide the prject name"
```
torbrenner, line 6: typo: prject -> project

lib/add_bc_user.sh
```diff
@@ -1,15 +1,29 @@
...
+fi
+
 read -p 'Username: ' bc_user
 read -sp 'Password: ' bc_password
 
-echo 
+echo
+
+
+bc=`docker run --rm -it httpd:latest htpasswd -nb $bc_user $bc_password`
```
torbrenner, line 21: Short question about the naming: what is the meaning behind "bc"?

torbrenner (changes requested):

lib/add_bc_user.sh
patrickskowronekdkfz: basic_user

lib/add_bc_user.sh
patrickskowronekdkfz: Fixed

PR #4, opened 2022-04-08
WIP: Feature/bd24

patrickskowronekdkfz:


dktk-fed/docker-compose.yml
```diff
@@ -24,9 +25,19 @@ services:
       - ../auth/:/auth
     extra_hosts:
       - "host.docker.internal:host-gateway"
+
   
+  ### Does need to know the outside proxy to connect central components
+  forward_proxy:
+    container_name: bridgehead-squid
+    image: ubuntu/squid
+    environment:
```
torbrenner, line 34: I would suggest to let read this from the environment like
```
http_proxy: ${http_proxy}
https_proxy: ${https_proxy}
```
We would need to ensure, that the system units will receive the proxy variables set by the user.

PR #5, opened 2022-04-08
Feature/bd 29

patrickskowronekdkfz:


configure-bridgehead.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+useradd --home-dir /srv/docker/bridgehead -g docker -N -u 317 -s /sbin/nologin
```
torbrenner, line 3: I would not set home-dir to /srv/docker/bridgehead. Use `-M` (don't create a home dir) option instead.
You would need to run `chown -R bridgehead:bridgehead /srv/docker/bridgehead` in the installation script after creating the user

torbrenner: don't know if there is a better default for `-s /sbin/nologin`  to remove the option for login

torbrenner: i wouldn't hardcode the userid to such a low number. Maybe we can find a better default

torbrenner: How do you ensure, that this user is named "bridgehead"?

lib/systemd/bridgehead@.service
```diff
@@ -2,14 +2,16 @@
...
+User=bridgehead
 Restart=always
 RestartSec=30
 
 WorkingDirectory=/srv/docker/bridgehead/
 
-ExecStartPre=/bin/bash -c '`which docker-compose` -f %i/docker-compose.yml --env-file site-config/%i.env down'
-ExecStart=/bin/bash -c '`which docker-compose` -f %i/docker-compose.yml --env-file site-config/%i.env up'
-ExecStop=/bin/bash -c '`which docker-compose` -f %i/docker-compose.yml --env-file site-config/%i.env down'
+ExecStartPre=/srv/docker/bridgehead/stop-bridgehead.sh %i
```
torbrenner, line 11: check if full path really is needed when working directory is set properly

lib/systemd/bridgehead@.service
```diff
@@ -2,14 +2,16 @@
...
+ExecStart=/srv/docker/bridgehead/start-bridgehead.sh %i
+RemainAfterExit=true
```
torbrenner, line 13: Instead of `RemainAfterExit=true` you could also ensure that `start-bridgehead.sh` calls `docker-compose logs` at the end. This would have the side effect, that all logs of the containers would be searchable with `journalctl -u`.
On the otherhand you could change the type of this systemd unit to `Oneshot`.

torbrenner (changes requested):

PR #20, opened 2022-10-05
Updated Blaze from v0.17 -> v0.18

torbrenner:
This fixes the issue with incorrect specimen response. The blaze data should be persisted as volume blaze-data is configured, so no reimport necessary. 
After merge all bridgeheads should pull the update at the start of the next hour.


alexanderkiel (approved):
The volume is ok.

PR #28, opened 2022-10-26
add port quick solution for ssl cert verification with portnumber; ge…

PierreDelpy:


ccp/nngm-setup.sh
```diff
@@ -7,4 +7,5 @@ function nngmSetup() {
 	fi
 }
 
-CONNECTOR_POSTGRES_PASSWORD="$(cat /proc/sys/kernel/random/uuid | sed 's/[-]//g' | head -c 20)"
+#CONNECTOR_POSTGRES_PASSWORD="$(cat /proc/sys/kernel/random/uuid | sed 's/[-]//g' | head -c 20)"
+CONNECTOR_POSTGRES_PASSWORD="$(echo -n /etc/bridgehead/pki/mannheim.priv.pem | sha256sum | head -c 20)"
```
lablans, line 11: This does not seem generic for all sites. 

Also, @TKussel, any other way to get a consistent password?

ccp/nngm-compose.yml
```diff
@@ -11,6 +11,8 @@ services:
       NNGM_CTS_APIKEY: ${NNGM_CTS_APIKEY}
       NNGM_CRYPTKEY: ${NNGM_CRYPTKEY}
     restart: always
+    ports:
+      - "8080:8080"
```
lablans, line 15: Do we need this open port? How about labels for traefik instead?

lablans (changes requested):
I'm sorry I can't accept this as is.

PR #38, opened 2022-11-18
WIP: Feature/monitoring: Adding Bridgehead Monitoring

patrickskowronekdkfz:
This PR adds the Bridgehead Monitoring for CCP. Currently only the blaze is monitored. This will increase with automatic updates.


ccp/docker-compose.yml
```diff
@@ -104,7 +106,17 @@ services:
...
+  monitoring:
+    image: "samply/bridgehead-monitoring:main"
+    container_name: bridgehead-monitoring
+    environment: 
+      HOST: ${HOST}
+      SITE_NAME: ${SITE_NAME}
+      PROJECT: ${PROJECT}
+      BEAM_URL: http://beam-proxy:8081
+      PROXY_ID: monitoring.${PROXY_ID}
+      MONITORING_TARGET: ${MONITORING_TARGET}
```
lablans, line 118: Please derive the `MONITORING_TARGET` from the project (you can use `$PROJECT/vars` file) rather than a site-specific config. For the simple reason that it's not site specific.

lablans (changes requested):
Will continue review once this is fixed. Thanks!

PR #43, opened 2022-11-25
refactor: Removed Debug Logging from Beam Proxy

torbrenner:
Some Sites reported issues with large log files. Due to this, i would first reduce the log level of the beam proxy.


patrickskowronekdkfz (approved):
I don't why we should keep it for now

PR #57, opened 2023-02-08
Added Module for Id-Management

torbrenner:
This adds all necessary scripts for deploying all local components of the CCP-IT id-management to the bridgehead.
The feature is activated by setting the `IDMANAGER_UPLOAD_APIKEY` configuration flag. For more information about this module, refer to the new id-management.md.

Because the database of the patientlist should always be reconstructable, I also added automated backups for the database. The automated backups are enabled through setting the `BACKUP_DIRECTORY` configuration.

At the moment I didn't add any monitoring checks for the id-management but will include them in a future PR.


PierreDelpy (approved):
Looks great!!

PR #60, opened 2023-02-22
During install, check if running in WSL and if systemd is present

TKussel:


lib/prepare-system.sh
```diff
@@ -5,6 +5,15 @@ source lib/functions.sh
 
 log "INFO" "Preparing your system for bridgehead installation ..."
 
+# Check, if running in WSL
+if [[ $(grep -i Microsoft /proc/version) ]]; then
+    # Check, if systemd is available
+    if [ $(systemctl is-system-running) -eq "offline" ]; then
```
torbrenner, line 11: Tested it in the commandline using bash. Seems like Bash doesn't like -eq with strings.

lablans: "" missing on the left side.

lablans, line 11:
```suggestion
    if [ "$(systemctl is-system-running)" == "offline" ]; then
```

lib/prepare-system.sh
```diff
@@ -5,6 +5,15 @@ source lib/functions.sh
...
+        log "ERROR" "It seems, that you have no active systemd environment in your WSL. Please follow the guide in https://devblogs.microsoft.com/commandline/systemd-support-is-now-available-in-wsl/"
```
lablans, line 12:
```suggestion
        log "ERROR" "It seems you have no active systemd environment in your WSL environment. Please follow the guide in https://devblogs.microsoft.com/commandline/systemd-support-is-now-available-in-wsl/"
```

lib/prepare-system.sh
```diff
@@ -5,6 +5,15 @@ source lib/functions.sh
 
 log "INFO" "Preparing your system for bridgehead installation ..."
 
+# Check, if running in WSL
+if [[ $(grep -i Microsoft /proc/version) ]]; then
```
lablans, line 9: What will happen if this has error code != 0?

lablans (changes requested):
please see comments in source code

lib/prepare-system.sh
TKussel: IIRC, `==` only works correctly with `[[ foo ]]`, no?

lib/prepare-system.sh
TKussel: According to the grep documentation, grep returns only 0, 1, and 2. 0 and 1 have the exact semantics we want in this comparison, 2 indicates an error and would lead to an exit(1) of the install script. While exiting is certainly a sensible approach (e.g. on Macs, that don't implement /proc/*), do you think we need more error handling here, @lablans?

lib/prepare-system.sh
lablans: Bash never ceases to surprise but this should work with single brackets, too.

lib/prepare-system.sh
lablans: My concern is that the whole bash script might crash if a non-zero return value is given here (as requested by the `-e` flag). Again, I don't know about bash to say for sure, just wanted to prevent an error. If you are sure this won't trigger `-e`, go ahead with the merge.

lablans (approved):

torbrenner (approved):

PR #64, opened 2023-03-08
Allow to automate installation

lablans:
Allow to provide the only two variables that are requested by the BK admin on installation via env variables.

Please confirm:
- [ ] Without new env variables: Installation works as before
- [ ] Alternatively, you can set either `site` or `access_token` for the `install` command and the installation won't prompt you anymore.


PierreDelpy:
@TKussel check via VM
@PierreDelpy I'll try it too

lablans:
@TKussel and @enola-dkfz , friendly reminder

PR #65, opened 2023-03-08
Dont delete all docker images

lablans:
Only clean up docker images if bridgehead is actually running.

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -34,7 +34,7 @@ checkOwner(){
 }
 
 printUsage() {
-	echo "Usage: bridgehead start|stop|update|install|uninstall|enroll PROJECTNAME"
+	echo "Usage: bridgehead start|stop|is-running|update|install|uninstall|enroll PROJECTNAME"
 	echo "PROJECTNAME should be one of ccp|bbmri"
 }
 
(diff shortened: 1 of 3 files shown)


lib/functions.sh
```diff
@@ -169,6 +169,17 @@ function retry {
   return 0
 }
 
+function bk_is_running {
+	detectCompose
+	RUNNING="$($COMPOSE -p bridgehead-$PROJECT -f ./$PROJECT/docker-compose.yml $OVERRIDE ps -q)"
+	NUMBEROFRUNNING=$(echo "$RUNNING" | wc -l)
+	if [ $NUMBEROFRUNNING -ge 2 ]; then
```
torbrenner, line 176: why greater equals than 2? Shouldn't we also accept 1 running service?

lib/update-bridgehead.sh
```diff
@@ -4,10 +4,15 @@ source lib/functions.sh
 AUTO_HOUSEKEEPING=${AUTO_HOUSEKEEPING:-true}
 
 if [ "$AUTO_HOUSEKEEPING" == "true" ]; then
-	A="Performing automatic maintenance: Cleaning docker images."
+	A="Performing automatic maintenance: "
+	if bk_is_running; then
+		A="$A Cleaning docker images."
+		docker system prune -a -f
```
PierreDelpy, line 10: sollen wir auch gleich ein volume prune ergänzen? (für ungenutzte volumes)

lib/functions.sh
lablans: That's the way `wc -l` works -- and there are no Bridgeheads with <2 services

lib/update-bridgehead.sh
lablans: Not sure (requires testing). Please make another PR for that

PR #72, opened 2023-03-29
Feature/teiler

djuarezgf:


djuarezgf:
Emil should create the following images:
docker.verbis.dkfz.de/cache/samply/teiler-root-config:latest
docker.verbis.dkfz.de/cache/samply/teiler-ui:latest

djuarezgf:
@torbrenner : I would change all the postgres-15.1-alpine through postgres-15-alpine. What do you think? (Then we have to add it to the cache of Harbor)

ccp/modules/exporter-setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ -n "$ENABLE_EXPORTER" ];then
```
lablans, line 3: Where is this variable defined? I can't find `ENABLE_EXPORTER` via Ctrl+F

ccp/modules/exporter-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+      POSTGRES_DB: "exporter"
+    volumes:
+      - "bridgehead-exporter-db:/var/lib/postgresql/data"
+
+
+volumes:
+  bridgehead-exporter-db:
+    name: "bridgehead-exporter-db"
+  bridgehead-exporter:
+    name: "bridgehead-exporter"
```
lablans, line 47: Please check if you can do without these volumes.

ccp/modules/exporter-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+      - "traefik.enable=true"
+      - "traefik.http.routers.exporter_ccp.rule=PathPrefix(`/ccp-exporter`)"
+      - "traefik.http.services.exporter_ccp.loadbalancer.server.port=8092"
+      - "traefik.http.routers.exporter_ccp.tls=true"
+      # TODO: Add basic auth in Teiler UI to access exporter?
+      #- "traefik.http.routers.exporter_ccp.middlewares=auth"
+      - "traefik.http.middlewares.exporter_ccp_strip.stripprefix.prefixes=/ccp-exporter"
+      - "traefik.http.routers.exporter_ccp.middlewares=exporter_ccp_strip"
+    volumes:
+      - "bridgehead-exporter:/app/exporter-files"
```
lablans, line 30: Please check if you can do without this volume.

ccp/modules/exporter-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  exporter-db:
+    image: postgres:15.1-alpine
+    container_name: bridgehead-ccp-exporter-db
+    environment:
+      POSTGRES_USER: "exporter"
+      POSTGRES_PASSWORD: "${EXPORTER_DB_PASSWORD}" # Set in exporter-setup.sh
+      POSTGRES_DB: "exporter"
+    volumes:
+      - "bridgehead-exporter-db:/var/lib/postgresql/data"
```
lablans, line 40: Please check if you can do without this volume.

ccp/modules/login-setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [ "$ENABLE_LOGIN" == true ];then
+  log INFO "Login setup detected -- will start Login services."
+  OVERRIDE+=" -f ./$PROJECT/modules/login-compose.yml"
+fi
```
lablans, line 6: Where is this variable set?

ccp/modules/login-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+      POSTGRES_DB: "keycloak"
+    volumes:
+      - "bridgehead-login-db:/var/lib/postgresql/data"
+
+  login:
+    image: docker.verbis.dkfz.de/ccp/dktk-keycloak:latest
+    container_name: bridgehead-login
+    environment:
+      KEYCLOAK_ADMIN: "admin"
+      KEYCLOAK_ADMIN_PASSWORD: "${KEYCLOAK_ADMIN_PASSWORD}"
```
lablans, line 20: Where is this variable defined?

ccp/modules/mtba-compose.yml
```diff
@@ -19,18 +19,15 @@ services:
...
-    labels:
-      - "traefik.enable=true"
-      - "traefik.http.routers.mtba.rule=PathPrefix(`/`)"
-      - "traefik.http.services.mtba.loadbalancer.server.port=80"
-      - "traefik.http.routers.mtba.tls=true"
+#    labels:
+#      - "traefik.enable=true"
+#      - "traefik.http.routers.mtba.rule=PathPrefix(`/mtba`)"
+#      - "traefik.http.services.mtba.loadbalancer.server.port=80"
+#      - "traefik.http.routers.mtba.tls=true"
```
lablans, line 26: Why has this been commented out?

ccp/modules/teiler-ui-setup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+if [ "$ENABLE_TEILER" == true ];then
```
lablans, line 3: Where is this variable defined?

lablans (changes requested):
It's a little hard to review since this PR also makes changes to other projects (e.g. nNGM, MTBA). I support the idea of moving everything into `modules/` but there might be changes in there the GitHub UI is not showing.

==> Could we have that in a separate PR?

ccp/modules/exporter-setup.sh
djuarezgf: This should be defined in the ccp.conf file

ccp/modules/login-setup.sh
djuarezgf: This should be defined in the ccp.conf file

ccp/modules/login-compose.yml
djuarezgf: This should be defined in the ccp.conf file

ccp/modules/teiler-ui-setup.sh
djuarezgf: This should be defined in the ccp.conf file

ccp/modules/mtba-compose.yml
djuarezgf: Currently, we don't need that it is externally reachable. It will be needed when we implement a micro-frontend for the Teiler-UI.

ccp/modules/exporter-compose.yml
djuarezgf: If we remove these volumes, the exports cannot be downloaded after restarting the bridgehead. It must be taken into account, that the export files need to be at least temporally saved,  as the process works asynchronously: 1) Execution of a query is requested. 2) Export can be downloaded as soon as it is available. 

If an export file can be saved only until the bridgehead is restarted, I would suggest removing the volume "bridgehead-exporter".

Until we solve the problem of query persistence, we need a way of saving the queries. As soon as we have a final solution for that, I would suggest keeping the volume "bridgehead-exporter-db".

ccp/modules/exporter-compose.yml
djuarezgf: See the answer to the previous comment.

djuarezgf:
TODO: Minimize teilerui_realm_export.json in dktk-keycloak

torbrenner:
> @torbrenner : I would change all the postgres-15.1-alpine through postgres-15-alpine. What do you think? (Then we have to add it to the cache of Harbor)

I don't think we should do this. As far as I know, postgres will always ask for manual user interaction (migrating database or running pg-migrate or something) after a minor update. So if postgres would suddenly decide to release postgres-15.2, all the postgres databases at the sites would fail to start and following that also all bridgeheads. For now, I would recommend to stick to 15.1 and search for a solution to automate the postgres minor updates.

PR #81, opened 2023-05-16
add max-time for curl monitoring

PierreDelpy:
I need a mechanism to reduce waiting time for sites that have connection issues. Otherwise it takes multiple minutes for each bridgehead restart depending on the default timeout duration

### DIFF
--- a/lib/monitoring.sh
+++ b/lib/monitoring.sh
@@ -47,8 +47,8 @@ function hc_send(){
 
     if [ -n "$2" ]; then
         MSG="$2\n\nDocker stats:\n$UPTIME"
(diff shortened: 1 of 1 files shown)


lib/monitoring.sh
```diff
@@ -47,8 +47,8 @@ function hc_send(){
 
     if [ -n "$2" ]; then
         MSG="$2\n\nDocker stats:\n$UPTIME"
-        echo -e "$MSG" | https_proxy=$HTTPS_PROXY_URL curl -A "$USER_AGENT" -s -o /dev/null -X POST --data-binary @- "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
+        echo -e "$MSG" | https_proxy=$HTTPS_PROXY_URL curl --max-time 5 -A "$USER_AGENT" -s -o /dev/null -X POST --data-binary @- "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
```
patrickskowronekdkfz, line 50: Is 5 second the right time?

lablans: I think so :)

lablans (approved):

PR #83, opened 2023-05-16
Change blaze tag to latest

patrickskowronekdkfz:
Update blaze to the latest tag


TKussel (approved):
As discussed wit @alexanderkiel, this is not a breaking change.

PR #90, opened 2023-06-19
Replace local spot with samply/focus

torbrenner:


torbrenner:
changes already merged

PR #104, opened 2023-08-07
BBMRI-ERIC / GBN combined Bridgehead

lablans:
This allows to operate one single Bridgehead for use with the GBN & BBMRI-ERIC Brokers.


patrickskowronekdkfz (approved):
The PR looks good, however I would make more sense to make a real module with focus and beam-proxy. Nevertheless, this can be merged and dealt with later.

PR #105, opened 2023-08-08
Feature/custom basic auth

PierreDelpy:


torbrenner (approved):

patrickskowronekdkfz (commented):
I would prefer the naming of setting a user, since adding would implying you could a another user.

PierreDelpy:
> I would prefer the naming of setting a user, since adding would implying you could a another user.

Besides creating required users once by default, this method **especially** allows adding other/additional users

patrickskowronekdkfz (approved):

PR #107, opened 2023-08-14
Dont require beam enroll to run interactively

Threated:
This is redundant I am pretty sure and it prevents running `./bridgehead enroll` in a script.


TKussel (approved):
Tested, works perfectly fine without interactive tty

PR #127, opened 2023-10-05
Added proxy user + pw detection

patrickskowronekdkfz:
Currently the config allows to use the user and password of a proxy as var or encoded in the url. The added function provides a possibility to cover both use cases

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -3,7 +3,7 @@
 detectCompose() {
 	if [[ "$(docker compose version 2>/dev/null)" == *"Docker Compose version"* ]]; then
 		COMPOSE="docker compose"
-	else
+	e
 		COMPOSE="docker-compose"
 		# This is intended to fail on startup in the next prereq check.
 	fi
@@ -15,8 +15,13 @@ setupProxy() {
 		PROTO="$(echo $HTTP_PROXY_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')"
 		URL="$(echo ${HTTP_PROXY_URL/$PROTO/})"
 		PROXY="$(echo $PROTO$HTTP_PROXY_USERNAME:$HTTP_PROXY_PASSWORD@$URL)"
+
+		SECURE_PROTO="$(echo $HTTPS_PROXY_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')"
+		SECURE_URL="$(echo ${HTTPS_PROXY_URL/$SECURE_PROTO/})"
+		SECURE_PROXY="$(echo $SECURE_PROTO$HTTP_PROXY_USERNAME:$HTTP_PROXY_PASSWORD@$SECURE_URL)"
 	else
 		PROXY=$HTTP_PROXY_URL
+		SECURE_PROXY=$HTTPS_PROXY_URL
 	fi
 }
 
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -16,11 +16,11 @@ setupProxy() {
 	local http="no"
 	local https="no"
 	if [ $HTTPS_PROXY_URL ]; then
-		local hostport=$(echo $HTTPS_PROXY_URL | sed -e s,$proto,,g | cut -d/ -f1)
+		local hostport=$(echo $HTTPS_PROXY_URL | sed -e "s,$proto,,g" | cut -d/ -f1)
 		HTTPS_PROXY_HOST="$(echo $hostport | sed -e 's,:.*,,g')"
 		HTTPS_PROXY_PORT="$(echo $hostport | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')"
 		if [[ ! -z "$HTTPS_PROXY_USERNAME" && ! -z "$HTTPS_PROXY_PASSWORD" ]]; then
-			local proto="$(echo $HTTPS_PROXY_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')"
+			local proto="$(echo $HTTPS_PROXY_URL | grep :// | sed -e 's,^\(.*://\).*,\1,g')"
 			local fqdn="$(echo ${HTTPS_PROXY_URL/$proto/})"
 			HTTPS_PROXY_FULL_URL="$(echo $proto$HTTPS_PROXY_USERNAME:$HTTPS_PROXY_PASSWORD@$fqdn)"
 
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -16,6 +16,8 @@ setupProxy() {
 	local http="no"
 	local https="no"
 	if [ $HTTPS_PROXY_URL ]; then
+		local proto="$(echo $HTTPS_PROXY_URL | grep :// | sed -e 's,^\(.*://\).*,\1,g')"
+		local fqdn="$(echo ${HTTPS_PROXY_URL/$proto/})"
 		local hostport=$(echo $HTTPS_PROXY_URL | sed -e "s,$proto,,g" | cut -d/ -f1)
 		HTTPS_PROXY_HOST="$(echo $hostport | sed -e 's,:.*,,g')"
 		HTTPS_PROXY_PORT="$(echo $hostport | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')"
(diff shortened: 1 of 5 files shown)


lib/update-bridgehead.sh
```diff
@@ -55,7 +55,7 @@ for DIR in /etc/bridgehead $(pwd); do
     OUT=$(retry 5 git -C $DIR fetch 2>&1 && retry 5 git -C $DIR pull 2>&1)
   else
     log "INFO" "Git is using proxy ${HTTP_PROXY_URL} from ${CONFFILE}"
-    OUT=$(retry 5 git -c http.proxy=$HTTP_PROXY_URL -c https.proxy=$HTTPS_PROXY_URL -C $DIR fetch 2>&1 && retry 5 git -c http.proxy=$HTTP_PROXY_URL -c https.proxy=$HTTPS_PROXY_URL -C $DIR pull 2>&1)
+    OUT=$(retry 5 git -c http.proxy=$PROXY -c https.proxy=$PROXY -C $DIR fetch 2>&1 && retry 5 git -c http.proxy=$PROXY -c https.proxy=$PROXY -C $DIR pull 2>&1)
```
torbrenner, line 58: You are mixing http and https proxy. This will lead to https calls beeing send to the http proxy, which could matter if the sites have two different proxies for that. I think we will need to have the code above create the PROXY variable for http and a SECURE_PROXY variable for https

torbrenner (changes requested):

lib/functions.sh
```diff
@@ -3,12 +3,28 @@
 detectCompose() {
 	if [[ "$(docker compose version 2>/dev/null)" == *"Docker Compose version"* ]]; then
 		COMPOSE="docker compose"
-	else
+	e
```
lablans, line 6: Even if this works, please don't do that

lib/functions.sh
```diff
@@ -3,12 +3,28 @@
...
+		PROXY="$(echo $PROTO$HTTP_PROXY_USERNAME:$HTTP_PROXY_PASSWORD@$URL)"
+
+		SECURE_PROTO="$(echo $HTTPS_PROXY_URL | grep :// | sed -e's,^\(.*://\).*,\1,g')"
+		SECURE_URL="$(echo ${HTTPS_PROXY_URL/$SECURE_PROTO/})"
+		SECURE_PROXY="$(echo $SECURE_PROTO$HTTP_PROXY_USERNAME:$HTTP_PROXY_PASSWORD@$SECURE_URL)"
+	else
+		PROXY=$HTTP_PROXY_URL
+		SECURE_PROXY=$HTTPS_PROXY_URL
+	fi
+}
```
lablans, line 26: Please consider other variable names, these are quite ambiguous.

lablans (changes requested):

lib/functions.sh
patrickskowronekdkfz: Fixed. I don't recall ever changing this var.

lib/prerequisites.sh
```diff
@@ -68,7 +68,7 @@ source /etc/bridgehead/${PROJECT}.conf
 source ${PROJECT}/vars
 
 set +e
-SERVERTIME="$(https_proxy=$HTTPS_PROXY_URL curl -m 5 -s -I $BROKER_URL_FOR_PREREQ 2>&1 | grep -i -e '^Date: ' | sed -e 's/^Date: //i')"
+SERVERTIME="$(https_proxy=$HTTPS_PROXY_FULL_URL curl -m 5 -s -I $BROKER_URL_FOR_PREREQ 2>&1 | grep -i -e '^Date: ' | sed -e 's/^Date: //i')"
```
TKussel, line 71:
```suggestion
SERVERTIME="$(http_proxy=$HTTP_PROXY_FULL_URL https_proxy=$HTTPS_PROXY_FULL_URL curl -m 5 -s -I $BROKER_URL_FOR_PREREQ 2>&1 | grep -i -e '^Date: ' | sed -e 's/^Date: //i')"
```
The broker is not required to run under HTTPS

lib/functions.sh
TKussel, line 1: This change does not pick up the case of an `HTTP[S]?_PROXY_USERNAME` without password. I don't know if that's legal in HTTP, though.

lib/monitoring.sh
```diff
@@ -47,8 +47,8 @@ function hc_send(){
 
     if [ -n "$2" ]; then
         MSG="$2\n\nDocker stats:\n$UPTIME"
-        echo -e "$MSG" | https_proxy=$HTTPS_PROXY_URL curl --max-time 5 -A "$USER_AGENT" -s -o /dev/null -X POST --data-binary @- "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
+        echo -e "$MSG" | https_proxy=$HTTPS_PROXY_FULL_URL curl --max-time 5 -A "$USER_AGENT" -s -o /dev/null -X POST --data-binary @- "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
```
TKussel, line 50:
```suggestion
        echo -e "$MSG" | http_proxy=$HTTP_PROXY_FULL_URL https_proxy=$HTTPS_PROXY_FULL_URL curl --max-time 5 -A "$USER_AGENT" -s -o /dev/null -X POST --data-binary @- "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
```
While our healthcheck is served via HTTPS, this is not required.

lib/monitoring.sh
```diff
@@ -47,8 +47,8 @@ function hc_send(){
...
     else
-        https_proxy=$HTTPS_PROXY_URL curl --max-time 5 -A "$USER_AGENT" -s -o /dev/null "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
+        https_proxy=$HTTPS_PROXY_FULL_URL curl --max-time 5 -A "$USER_AGENT" -s -o /dev/null "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
```
TKussel, line 52:
```suggestion
       http_proxy=$HTTP_PROXY_FULL_URL  https_proxy=$HTTPS_PROXY_FULL_URL curl --max-time 5 -A "$USER_AGENT" -s -o /dev/null "$HCURL"/"$1" || log WARN "Monitoring failed: Unable to send data to $HCURL/$1"
```

lib/update-bridgehead.sh
```diff
@@ -50,12 +50,11 @@ for DIR in /etc/bridgehead $(pwd); do
     git -C $DIR config credential.helper "$CREDHELPER"
   fi
   old_git_hash="$(git -C $DIR rev-parse --verify HEAD)"
-  if [ -z "$HTTP_PROXY_URL" ]; then
-    log "INFO" "Git is using no proxy!"
+  if [ -z "$HTTP_PROXY_FULL_URL" ]; then
```
TKussel, line 53: This is wrong, if `HTTPS_PROXY_FULL_URL` is set, but `HTTP_PROXY_FULL_URL` is not!

lib/update-bridgehead.sh
```diff
@@ -50,12 +50,11 @@ for DIR in /etc/bridgehead $(pwd); do
     git -C $DIR config credential.helper "$CREDHELPER"
   fi
   old_git_hash="$(git -C $DIR rev-parse --verify HEAD)"
-  if [ -z "$HTTP_PROXY_URL" ]; then
-    log "INFO" "Git is using no proxy!"
```
TKussel, line 54: Why did you remove that?

TKussel (changes requested):

lib/functions.sh
lablans: agreed -- it would be technically legal but so uncommon that it doesn't matter

lib/monitoring.sh
lablans: Service is in fact required to run under https, cf. data protection concept(s)

lib/prerequisites.sh
lablans: Service is in fact required to run under https, cf. data protection concept(s)

lib/monitoring.sh
lablans: Service is in fact required to run under https, cf. data protection concept(s)

lib/update-bridgehead.sh
lablans: Thanks, fixed in 9fc8564e4ef280403deb472a40eda255e02d740e

lib/update-bridgehead.sh
lablans: thanks, fixed in 9fc8564e4ef280403deb472a40eda255e02d740e

TKussel:
I commented out the usage of an HTTP_PROXY as it is forbidden by our data protection concepts

TKussel (approved):

lablans (approved):

PR #128, opened 2023-10-06
Fix script for the creation of persistent keys

TKussel:
The creation of the permanent keys in the bridgehead installer was broken, as openssl expects the input to be a (sha1 or md5) hash. This is fixed now. 

Note: This PR breaks the keys of existing bridgehead installations. Only merge after creating a migration plan.


lablans (changes requested):

TKussel:
With further inspection, `PATIENTLIST_POSTGRES_PASSWORD` works as it is and `DNPM_APPLICATION_SECRET` is unused and will be removed. Therefore, I close the PR and document OpenSSL's behavior for future occurrences.

PR #132, opened 2023-10-25
git requires http proxy config even vor https connections

TKussel:
Even if the repository URL uses HTTPS, git uses the http.proxy configuration if the *proxy* is reachable via HTTP.


patrickskowronekdkfz (approved):
I hope that all sites use the same address for http and https proxy.

lablans (approved):

PR #134, opened 2023-11-02
Enable obfuscation for ccp queries

TKussel:
Enable statistical obfuscation for CCP query results. Chosen Parameters:

* Epsilon = 0.28
* Sensitivity Patients/Specimen/Diagnosis/Procedures/Medication Statements= 1/20/3/1.7/2.1
* Don't obfuscate real zeros
* Display 10 for 0 < count < 10
* Round to the next 10s

Required for merging:

- [x] Handle DKTK MeasureReport in Focus
- [x] Test in Mannheim
- [x] Go-ahead from CCP office

### DIFF
--- a/ccp/docker-compose.yml
+++ b/ccp/docker-compose.yml
@@ -28,7 +28,7 @@ services:
       BLAZE_URL: "http://bridgehead-ccp-blaze:8080/fhir/"
       BEAM_PROXY_URL: http://beam-proxy:8081
       RETRY_COUNT: ${FOCUS_RETRY_COUNT}
-      OBFUSCATE: "no"
+      EPSILON: 0.28
     depends_on:
       - "beam-proxy"
       - "blaze"


enola-dkfz:
Focus is not ready because MeasureReport in DKTK differs from the one in BBMRI.

TKussel:
> Focus is not ready because MeasureReport in DKTK differs from the one in BBMRI.

Thanks, I have added this precondition to the PR description

enola-dkfz (approved):
Best to merge after Focus develop is merged into main

patrickskowronekdkfz (changes requested):
Currently the ccp focus is also used by the exliquid dashboard, which does not want obfuscation. My proposal would be to add another focus for this project.

TKussel:
What about this: we include the obfuscation state as a variable, that you can override in the exliquid setup?

enola-dkfz:
> What about this: we include the obfuscation state as a variable, that you can override in the exliquid setup?

In which application would that be done? If it's a parameter sent by Lens, that's a security issue. 

Because project/query language is already intended to be sent in metadata of the query, inserted by the Very Large Telescope, we could obfuscate, or not, depending on the project.

enola-dkfz:
now the obfuscation is turnoffable in Focus based on the project
it is automatically turned off for the exliquid project
the beam tasks coming from exliquid dashboard should have this metadata
"metadata":{"project":"exliquid"}

patrickskowronekdkfz (approved):

PR #136, opened 2023-11-30
Bugfix: MTBA 1.0.0

djuarezgf:


ccp/modules/mtba-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   mtba:
-    image: docker.verbis.dkfz.de/cache/samply/mtba:develop
+    image: docker.verbis.dkfz.de/cache/samply/mtba:1.0.0
```
lablans, line 5:
```suggestion
    image: docker.verbis.dkfz.de/cache/samply/mtba:latest
```

Would it be okay to use `latest` tag?

PR #139, opened 2023-12-15
Cleanup dnpm connect module

TKussel:
* Delete remains of DKTK Broker migration
* Add docker host alias to dnpm-connect
* Remove `DNPM_APPLICATION_SECRET`, that is only required in a different module


patrickskowronekdkfz (commented):
There @TKussel I'm at bit confused about the structure of the setup files. The setup will be called in minimal and the respective project. Is this still necessary? Could it not be run completely in minimal? 

Secondly, does one site use the bbmri bridgehead for dnpm. If so, does the bbmri priv.pem work with the ccp broker when they use the same cert?

TKussel:
The module under minimal is only called, if the project is minimal as well. Otherwise, only the project specific module is initialized.

Currently, no site is using a bbmri bridgehead. This would require an additional enrollment, but no additional private key. I assume, that this is mostly hypothetical. I foresee only "ccp" and "minimal" bridgeheads in dnpm.

patrickskowronekdkfz (approved):
Looks good. I would prefer to have not double all the scripts, but this is a future issue.

TKussel:
> Looks good. I would prefer to have not double all the scripts, but this is a future issue.

Thanks. I fully agree, let us tackle this issue in the new year!

PR #140, opened 2023-12-15
Feature/dnpm node

TKussel:
This adds a module to run the DNPM Node in the bridgehead by setting `ENABLE_DNPM_NODE` in the configuration.
Additionally, `DNPM_DATA_DIR` and `ZPM_SITE` need to be set as well.

Activating this, creates/appends a minimal/docker-compose.override.yml exposing the dashboard on a different (non-root) path.


patrickskowronekdkfz (changes requested):
The path "/dashboard" is already set by the dashboard of traefik, see here https://github.com/samply/bridgehead/blob/0b90cdb769cc53ca9b6a89b0ec4fd8cb5f79c4bd/minimal/docker-compose.yml#L19. Since nobody is using the dashboard right now, I would rather remove the traefik dashboard.

TKussel:
Good catch! Fixed in f66f2755d89cf0178ce201003bf999ac4e6ec6a7
I renamed the landing page override, as the landingpage will be replaced soon, and I did not want to break any existing traefik workflows.

patrickskowronekdkfz (approved):

PR #141, opened 2023-12-15
Add nngm module to minimal project

TKussel:
For running a combined DNPM and nNGM bridgehead it is required to use the nngm module out of the minimal project.

### DIFF
--- a/minimal/modules/nngm-compose.yml
+++ b/minimal/modules/nngm-compose.yml
@@ -1,0 +1,29 @@
+version: "3.7"
+volumes:
+  nngm-rest:
+
+services:
+  connector:
+    container_name: bridgehead-connector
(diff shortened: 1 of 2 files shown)


patrickskowronekdkfz (commented):
Maybe blocked. There are currently existing docker-compose.overrides, where I'm not quite sure if they will be included when moved to the minimal folder. I will check that.

TKussel:
> Maybe blocked. There are currently existing docker-compose.overrides, where I'm not quite sure if they will be included when moved to the minimal folder. I will check that.

Note: The module will not be moved, but is copied over. We will have sites using the ccp project, as well as minimal project. Sites operating a ccp bridgehead with nngm are not affected by this PR.

patrickskowronekdkfz (approved):

PR #142, opened 2024-01-05
Added test root certs and logic for beam to use test brokers

patrickskowronekdkfz:
This PR will enable bridgehead admin to switch to the test brokers.
The Implementation is adequate and maybe could be better.


bbmri/modules/eric-setup.sh
```diff
@@ -6,6 +6,13 @@ if [ "${ENABLE_ERIC}" == "true" ]; then
 
 	# Set required variables
 	ERIC_BROKER_ID=broker.bbmri.samply.de
```
enola-dkfz, line 8: In case it's test env you set them twice, first for production, and then for test
do it in if-then-else instead

bbmri/modules/gbn-setup.sh
```diff
@@ -6,6 +6,12 @@ if [ "${ENABLE_GBN}" == "true" ]; then
 
 	# Set required variables
 	GBN_BROKER_ID=broker.bbmri.de
```
enola-dkfz, line 8: In case it's test env you set them twice, first for production, and then for test
do it in if-then-else instead

bbmri/modules/eric-setup.sh
```diff
@@ -6,6 +6,13 @@ if [ "${ENABLE_ERIC}" == "true" ]; then
...
+	ERIC_ROOT_CERT=eric
+
+	if [ "{$ENABLE_TEST}" == "true" ]; then
```
lablans, line 11:
```suggestion
	if [ "${ENABLE_TEST}" == "true" ]; then
```

Is this tested? Because with this typo it shouldn't work

lablans (commented):
I'm making this into a more generic variable. Will provide PR shortly.

lablans (approved):
Here's the promised commit.

Please make 100% sure to **test this** (don't forget to test with DKTK), then GO for merge.

PierreDelpy (approved):
DKTK tested successfully

PR #144, opened 2024-01-25
Recommendation for standalone Bridgehead

DavidCroftDKFZ:
Requested by Zdenka

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ This repository is the starting point for any information and tools you will nee
 
 ## Requirements
 
-We recommend a dedicated VM for the Bridgehead, with nothing else running on it.
+To guarantee a smooth operation of the Bridgehead, we recommend a dedicated VM for the Bridgehead, with no other applications running on it.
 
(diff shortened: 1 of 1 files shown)


patrickskowronekdkfz (changes requested):
We should advise what is the minimum of version of git, curl and docker.

README.md
```diff
@@ -34,6 +34,10 @@ This repository is the starting point for any information and tools you will nee
 
 ## Requirements
 
+We recommend a dedicated VM for the Bridgehead, with nothing else running on it.
+
+It may to be possible to run other apps on the same server, if they don't share common ports with the Bridgehead, and if they do not take up resources that the Bridgehead needs, like RAM. The Bridgehead may also have problems if other applications need older versions of git, Docker or curl.
```
TKussel, line 39:
```suggestion
To guarantee a smooth operation of the Bridgehead, we recommend a dedicated VM for the Bridgehead, with no other applications running on it.

It may to be possible to run other apps on the same server, if they don't share common ports with the Bridgehead, and if they do not take up resources that the Bridgehead needs, like RAM. The Bridgehead may also run into issues if other applications need incompatible versions of git, Docker, curl, or other dependencies.
```

TKussel (approved):
LGTM, my suggestion for change is exactly that: a suggestion.

PR #73, opened 2023-04-17
Teiler, Exporter/Reporter, DataSHIELD, MTBA

djuarezgf:
This change is [Reviewable](https://reviewable.io/reviews/samply/bridgehead/73)

PR #145, opened 2024-01-26
Remove local Keycloak Installation

torbrenner:
- Removes the need for local installation of Keycloak
- Adds functions for automated secret_sync (@Threated, maybe you can explain that better)
- Adds helper functions for creating secrets
- Renames some variables and separates secrets that shouldn't be the same (e.g. OPAL_EXPORTER_PASSWORD is no longer the same as LDM_AUTH)


torbrenner:
It seems like this should be merged into feature/datashield before merging #73, as it already addresses some complaints. See description for more information.
If it is possible, please also remove all files for the login module, and [ccp/modules/teiler-ui-compose.yml](https://github.com/samply/bridgehead/pull/145/files#diff-b6e36aa0a46e593ac038b48b45a9e1ca174ec01076bcb42bc836a7c884b69ab8) as it is empty

djuarezgf:
We can merge it after Adrian Vajda has moved the bridgehead's configuration of the central keycloak to a new realm (instead of the master realm)

Threated (approved):

patrickskowronekdkfz (approved):
There are some lines which could be improved, but mostly it is fine

PR #146, opened 2024-01-30
Fix image updates for image names with vars

Threated:
This now uses `docker compose config --images` to gather all used images instead of trying to find the image names with a regex.
This has the added benefit of removing comments and interpolating env vars for us.
This fixes the focus updates not getting pulled since #142 because they get their tag from env since that version.
The only concern here is that `docker-compose config` and `docker compose config` are not present on the docker versions at some sites.

### DIFF
--- a/lib/update-bridgehead.sh
+++ b/lib/update-bridgehead.sh
@@ -86,7 +86,7 @@ done
 # Check docker updates
 log "INFO" "Checking for updates to running docker images ..."
 docker_updated="false"
-for IMAGE in $(cat $PROJECT/docker-compose.yml ${OVERRIDE//-f/} minimal/docker-compose.yml | grep -v "^#" | grep "image:" | sed -e 's_^.*image: \(.*\).*$_\1_g; s_\"__g'); do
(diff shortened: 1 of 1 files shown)


lablans:
> The only concern here is that `docker-compose config` and `docker compose config` are not present on the docker versions at some sites.

Thanks for the important remark. This would break Bridgehead scripts at some sites. Can we have some kind of fallback?

Threated:
In `docker compose config` it was added 3years ago https://github.com/docker/compose/commit/45956c36fba5d50cc43312600dd848a2187dc0d8 which was released with v2.2.0
I was not able to find it in the old [docker/compose-cli](https://github.com/docker/compose-cli) repo so maybe it only works on the `docker-compose` versions that use `docker compose` under the hood.

PierreDelpy (approved):
works for DKTK

Threated:
This now uses compose config in a way that is supported by version 2.xx

PierreDelpy:
before merging, please solve this tag with git magic https://github.com/samply/bridgehead/blob/main/ccp/docker-compose.yml#L22

PR #148, opened 2024-02-09
Bump postgres version to 15.6

lablans:
- [x] Read Release Notes for 15.5 and 15.6
- [x] Try on existing BK if this migrates without issue.

PR #149, opened 2024-02-09
Externalize postgres version

lablans:


torbrenner (changes requested):
only merge after #148, the postgres image referenced there is not available.

djuarezgf:
15.6-alpine is already there: https://hub.docker.com/layers/library/postgres/15.6-alpine/images/sha256-9134f6c39b435bc9d83aa5ca34425082631ea081934c78075d2ef35116fd4077?context=explore

PR #150, opened 2024-02-14
Add `bridgehead logs` command

Threated:


patrickskowronekdkfz (commented):
Would journalctl not be better?

Threated:
journalctl always loses colors (maybe there is some flags to prevent that) and I like narrowing down which services to look at with:
```bash
./bridgehead logs ccp bridgehead-beam bridgehead-focus
```
Or using the other docker compose logs flags like `--since` and so on.
Some of this of course works with journalctl and grep as well but I dont like it.

patrickskowronekdkfz (approved):

lablans (approved):

PR #154, opened 2024-02-20
Optimize memory usage of Blaze

torbrenner:
This will adjust the memory used by blaze on the different servers. The bridgehead start script reads the available memory from /proc/meminfo and calculates the amount to give to blaze in Megabyte. With this, blaze should receive approxymately one half of the RAM available to the systems, as one quarter is used for each, the java heap size and the db cache.
I tested it locally in wsl, where the amount used by blaze was reduced from 4G to 1951m. It would be good if someone could test it on a server.
- [x] test in local environment (wsl)
- [x] test on server

### DIFF
--- a/bbmri/docker-compose.yml
+++ b/bbmri/docker-compose.yml
@@ -9,6 +9,7 @@ services:
     environment:
       BASE_URL: "http://bridgehead-bbmri-blaze:8080"
       JAVA_TOOL_OPTIONS: "-Xmx${BLAZE_MEMORY_CAP:-4096}m"
+      DB_RESOURCE_CACHE_SIZE: ${BLAZE_RESOURCE_CACHE_CAP:-2500000}
       DB_BLOCK_CACHE_SIZE: $BLAZE_MEMORY_CAP
       ENFORCE_REFERENTIAL_INTEGRITY: "false"
     volumes:
--- a/bbmri/docker-compose.yml
+++ b/bbmri/docker-compose.yml
@@ -8,8 +8,8 @@ services:
     container_name: bridgehead-bbmri-blaze
     environment:
       BASE_URL: "http://bridgehead-bbmri-blaze:8080"
-      JAVA_TOOL_OPTIONS: "-Xmx4g"
-      LOG_LEVEL: "debug"
+      JAVA_TOOL_OPTIONS: "-Xmx${BLAZE_MEMORY_CAP}m"
+      DB_BLOCK_CACHE_SIZE: $BLAZE_MEMORY_CAP
       ENFORCE_REFERENTIAL_INTEGRITY: "false"
     volumes:
       - "blaze-data:/app/data"
--- a/bbmri/docker-compose.yml
+++ b/bbmri/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     container_name: bridgehead-bbmri-blaze
     environment:
       BASE_URL: "http://bridgehead-bbmri-blaze:8080"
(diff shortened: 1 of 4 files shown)


lib/functions.sh
```diff
@@ -155,6 +155,16 @@ setHostname() {
 	fi
 }
 
+# blaze memory cap should be approximately a quarter of the system memory
+# the memory cap will be applied to both the java heap size and db clock cache
+setBlazeMemoryCap() {
+	if [ -z "$BLAZE_MEMORY_CAP" ]; then
+	   system_memory=$(grep MemTotal /proc/meminfo | grep -Po '\d+');
+	   system_memory_in_mb=$(("$system_memory"/1024));
```
TKussel, line 163: Depending on the /proc file system is always a bit tricky and potentially not portable. How about the following:
```suggestion
	   system_memory_in_mb=$(free -m | grep 'Mem:' | awk '{print $2}');
```

bbmri/docker-compose.yml
```diff
@@ -8,8 +8,8 @@ services:
     container_name: bridgehead-bbmri-blaze
     environment:
       BASE_URL: "http://bridgehead-bbmri-blaze:8080"
-      JAVA_TOOL_OPTIONS: "-Xmx4g"
-      LOG_LEVEL: "debug"
+      JAVA_TOOL_OPTIONS: "-Xmx${BLAZE_MEMORY_CAP}m"
```
TKussel, line 11: Maybe add a 4g limit as a default value?
```suggestion
      JAVA_TOOL_OPTIONS: "-Xmx${BLAZE_MEMORY_CAP:-4096}m"
```

TKussel (changes requested):

patrickskowronekdkfz:
Since we only set JAVA Heap and not the DB_BLOCK_CACHE_SIZE, your example might be faster in certain cases then with only 4GB of JAVA Heap. Maybe we should test this as well.

Anyways, Maybe it makes also sense to inform the User that the available RAM is not enough to have good performance.

bbmri/docker-compose.yml
```diff
@@ -8,8 +8,8 @@ services:
     container_name: bridgehead-bbmri-blaze
     environment:
       BASE_URL: "http://bridgehead-bbmri-blaze:8080"
-      JAVA_TOOL_OPTIONS: "-Xmx4g"
-      LOG_LEVEL: "debug"
```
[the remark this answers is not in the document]
torbrenner: Yes, i don't think we should use debug log level on production sites. As far as i know in the initial configuration we activated debug unintentional and removed it later for ccp.
But to be sure, we should discuss this change first.

bbmri/docker-compose.yml
```diff
@@ -8,8 +8,8 @@ services:
...
+      JAVA_TOOL_OPTIONS: "-Xmx${BLAZE_MEMORY_CAP:-4096}m"
+      DB_BLOCK_CACHE_SIZE: $BLAZE_MEMORY_CAP
```
patrickskowronekdkfz, line 12: DB_RESOURCE_CACHE_SIZE is missing which is also an important Value. We would need for 2500000 for 4GB RAM. However this value depends on the physical size of data and the amount of resources.

patrickskowronekdkfz (changes requested):
We would need to set DB_RESOURCE_CACHE_SIZE, according to https://github.com/samply/blaze/blob/master/docs/tuning-guide.md, however this would imply a default value which could make the performance worse.

patrickskowronekdkfz (approved):
I think now everything in it should work, we should test it at site and compare the performance.

PR #155, opened 2024-02-21
Add DNPM_NO_PROXY configuration option

TKussel:
Some sites need to add additional no_proxy targets to the dnpm-beam-connect service.
This PR introduces an optional `DNPM_NO_PROXY` config variable, which, if set, adds the string (with an prepended comma) to the already existing dnpm-beam-connect no_proxy environment.


torbrenner:
Why do we need a separate variable instead of just redefining NO_PROXY in the specific sites configuration? Is it likely, that the configuration will be the same for all sites dnpm sites?

TKussel:
This will not be the same for all dnpm sites and it should only affect one service, that is communicating purely within the institution's network. Hence, the "global" NO_PROXY env would do "too much".

torbrenner (approved):

PR #156, opened 2024-02-22
mount and process trusted certs in dnpm-beam-connect

TKussel:


Threated (approved):
Btw I pushed a commit to beam-connect which will now skip folders and files it fails to parse (will give a warning).

PR #185, opened 2024-05-21
switch focus of ccp to tag

patrickskowronekdkfz:
This will switch the ccp project to the focus tag depending of environment.


enola-dkfz (approved):
splendid 👯‍♂️

PR #186, opened 2024-06-05
feat: Add logs command for journalctl

Threated:
I added `-a` as a default option for journalctl as it preserves the colors and fixes an issue where docker logs would only show up as `[blob data x kb]`. I dont think it has any downside

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ You can either do this in a browser or with curl. If you visit the URL in the br
 curl -k https://localhost
 ```
 
-If you get errors when you do this, you can inspect the logs of your landing page container in order to determine what is going wrong. To do this you can use `./bridgehead docker-logs <Project> landing -f` to follow the logs of the container. This transaltes to a docker compose logs command meaning all the ususal docker logs flags work.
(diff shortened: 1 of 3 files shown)


lablans (approved):

README.md
```diff
@@ -235,7 +236,7 @@ You can either do this in a browser or with curl. If you visit the URL in the br
 curl -k https://localhost
 ```
 
-If you get errors when you do this, you need to use ```docker logs``` to examine your landing page container in order to determine what is going wrong.
+If you get errors when you do this, you can inspect the logs of your landing page container in order to determine what is going wrong. To do this you can use `./bridgehead docker-logs <Project> landing -f` to follow the logs of the container. This transaltes to a docker compose logs command meaning all the ususal docker logs flags work.
```
patrickskowronekdkfz, line 239: Why landingpage? It should be 
```suggestion
Should the landing page not show anything, you can inspect the logs of the containers to determine what is going wrong. To do this you can use `./bridgehead docker-logs <Project> -f` to follow the logs of the container. This transaltes to a docker compose logs command meaning all the ususal docker logs flags work.
```

patrickskowronekdkfz (changes requested):

PR #190, opened 2024-07-01
feature: upgrade to oBDS2FHIR

PierreDelpy:


ccp/modules/obds2fhir-rest-setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+function obds2fhirRestSetup() {
+  if [ -n "$ENABLE_OBDS2FHIR_REST" ]; then
+    log INFO "oBDS2FHIR-REST setup detected -- will start obds2fhir-rest API."
```
patrickskowronekdkfz, line 5:
```suggestion
    log INFO "oBDS2FHIR-REST setup detected -- will start obds2fhir-rest module."
```

ccp/modules/obds2fhir-rest-setup.sh
```diff
@@ -0,0 +1,13 @@
...
+    if [ ! -n "$IDMANAGER_LOCAL_PATIENTLIST_APIKEY" ]; then
+      log WARN "Missing ID-Management Module! Fix this by setting up ID Management:"
```
patrickskowronekdkfz, line 7: Please do an exit 1(I think), instead of starting everything and then crash it

patrickskowronekdkfz (changes requested):

ccp/modules/obds2fhir-rest-setup.sh
PierreDelpy: This was the configuration in the past. However, there are bridgeheads that do not have a running ID management on the same server. Therefore this is more of a "new feature"

patrickskowronekdkfz: With -n you check if the APIKEY is empty. If it is empty I think you want to exit instead of wait if it works.

We are doing it https://github.com/samply/bridgehead/blob/e28b125b9368af9906a41c1bab9ac907dc9e6e5a/ccp/modules/mtba-setup.sh#L6 for example

PierreDelpy: Sometimes there is no ID management running. In the past, bridgeheads that required "MTBA" or "ADT2FHIR-REST" then had to locally overwrite this "exit 1" command, in order to run the bridgehead with the necessary modules. 
Since local configuration changes are not desired (or worse than this propsal), I removed this hard requirement and left a warning.

patrickskowronekdkfz: Then please use ERROR instead

PierreDelpy: done and squashed to 1 commit

patrickskowronekdkfz (approved):

PR #195, opened 2024-07-23
add caching in focus

PierreDelpy:
~~This PR only works if the config Repo contains the queries_to_cache.conf~~ ~~https://git.verbis.dkfz.de/bridgehead-configurations/bridgehead-config-mannheim/-/blob/develop/queries_to_cache.conf?ref_type=heads~~ Updated to have everything directly in ccp directory (the 2 config lines represent an empty query and a prostata adeno-ca query)

This branch is checked out in Mannheim and the config is also committed (see link above; feel free to test in https://data.dktk.dkfz.de)
The reason why I suggest to code it like this (bridgehead fails/logs infinite warnings if the config is missing) is, that we should add this config to every site


enola-dkfz (approved):
This works, and the file can be left empty in BHs where no caching is needed.

ccp/docker-compose.yml
```diff
@@ -31,6 +31,9 @@ services:
       BEAM_PROXY_URL: http://beam-proxy:8081
       RETRY_COUNT: ${FOCUS_RETRY_COUNT}
       EPSILON: 0.28
+      QUERIES_TO_CACHE: '/queries_to_cache.conf'
+    volumes:
+      - /etc/bridgehead/queries_to_cache.conf:/queries_to_cache.conf
```
TKussel, line 36: This would result in the creation of a folder, if the file does not exist already. Can we find a backwards compatible way? An optimization such as caching should not cause operational issues.

TKussel (changes requested):

ccp/docker-compose.yml
PierreDelpy: Changed it to include the config file directly in the same repo and directory, better?

TKussel: This is on the same level as the ccp vars, right? I think this is absolutely fine!

TKussel (approved):

ccp/docker-compose.yml
```diff
@@ -31,6 +31,9 @@ services:
       BEAM_PROXY_URL: http://beam-proxy:8081
       RETRY_COUNT: ${FOCUS_RETRY_COUNT}
       EPSILON: 0.28
+      QUERIES_TO_CACHE: '/queries_to_cache.conf'
+    volumes:
+      - ./queries_to_cache.conf:/queries_to_cache.conf
```
Threated, line 36: I am a bit skeptical of the relativ path. If it works it works but in most places we hardcoded `/srv/docker/bridgehead`

PierreDelpy: thx, changed

enola-dkfz:
The newest approach would have all the sites cache results of the same queries
Do we want to let sites choose their own queries whose results are to be caches based on the types and diagnoses of samples/patients, slowness of their hardware, etc?

PierreDelpy:
> The newest approach would have all the sites cache results of the same queries Do we want to let sites choose their own queries whose results are to be caches based on the types and diagnoses of samples/patients, slowness of their hardware, etc?

I think it is fine if all ccp sites have the same cache (it is even better than different caches for the same project)

torbrenner (approved):

PR #196, opened 2024-07-29
Add new/additional dashboard backend

davidmscholz:


Threated (approved):

davidmscholz:
implementing some queries required for the Exliquid Dashboard logic and potentially for the Organoid Dashboard(s) in CQL turned out to be surprisingly hard.

The approach that should be explored with this dashboard backend is to copy FHIR Resources (as they are - no additional mapping applied) from a Blaze FHIR Search over to PostgreSQL into a JSONB column.

In addition to being able to query Blaze using CQL this should enable us to query PostgreSQL with all its JSON functions and operators (see https://www.postgresql.org/docs/current/functions-json.html). A whitelist of SQL queries is held in Focus in order to make sure that only aggregated data leaves the bridgehead(s).

ccp/modules/dashboard-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: "3.7"
+
+services:
+  fhir2sql:
+    depends_on:
+      - "dashboard-db"
```
TKussel, line 6: it depends on blaze as well, no?

ccp/modules/dashboard-setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
+
+if [ "$ENABLE_DASHBOARD" == true ]; then
```
TKussel, line 3: Please make this env more specific

TKussel (changes requested):

ccp/modules/dashboard-compose.yml
davidmscholz: true

ccp/modules/dashboard-setup.sh
davidmscholz: would $ENABLE_FHIR2SQL be okay?

ccp/modules/dashboard-compose.yml
davidmscholz: thanks for pointing this out. Connecting to blaze from fhir2sql turns out to be a bit more tricky than I initially expected. From what I understand fhir2sql would not be able to connect to the blaze container like this anyway because blaze is defined in a different docker-compose file and is thus in a separate network. Correct? 

One way to address this would be to create a shared network using the ```networks``` directive.

TKussel: Good question. I was under the impression, that only one docker network is created, as we `-f` all the docker compose files.

Threated: This work docker will literally use a yaml merge lib to merge all the docker files and pretend it's one docker file

TKussel (approved):

Threated:
Should we merge this into main or rather our new develop branch?

PR #197, opened 2024-07-29
In ENVIRONMENT=production, use main tag for Samply.Beam.

lablans:


Threated (approved):

patrickskowronekdkfz (approved):

bridgehead
```diff
@@ -80,13 +80,16 @@ loadVars() {
 	case "$ENVIRONMENT" in
 		"production")
 			export FOCUS_TAG=main
+			export BEAM_TAG=main
 			;;
 		"test")
 			export FOCUS_TAG=develop
+			export BEAM_TAG=develop
```
patrickskowronekdkfz, line 87: Can we set all tags to develop?

patrickskowronekdkfz (changes requested):
Please thin

bridgehead
lablans: If I understand you correctly, you'd like to switch **all** tags to `develop` in case of a test environment. That makes sense but the tags are not the same for all images. Also, we'd like to make it a conscious choice to "onboard" a component into the "environment setting", following a short discussion with the components maintainer.

So YES, we can onboard more/all components into the environment BUT we'd need to do it one by one. So if you need a component onboarded, please create a separate issue.

patrickskowronekdkfz: Yes, do did understand me correctly. I think it would make sense to have something like 
`export IMAGE_TAG=develop` and use it where we can use it right now for beam and focus. I agree we would change it later at the other components

PR #198, opened 2024-07-29
Switch ID-Management to Keycloak from Samply.Auth

torbrenner:
This PR introduces a new way for Authentication in CCP ID-Management. Instead of Samply Auth, we use Keycloak as a provider. For this we added oauth2-proxy to handle the authentication and refactored the implementation in MagicPL to only do the check of the users access token and necessary groups. Unfortunately, i couldn't get it to work that we only have one installation of the oauth2-proxy running in the bridgehead. DataSHIELD still needs it's own, as both idm and datashield require a specific group for access and i was not able to separate this only in the labels.
Things we need to do before merging:
- [ ] Code Review for MagicPL
- [x] Test on Test Server (i provided an installation of this on our test bridgehead)
- [x] Create Clients for Bridgeheads in Keycloak


Threated (approved):
Maybe add a depends on oauth proxy to magic pl although it will probably start waaay slower than the oauth proxy anyway

patrickskowronekdkfz (approved):

PR #200, opened 2024-08-01
Reduce bridgehead update interval to once a day at 6am

TKussel:
Currently, the bridgehead updates every full hour. As we have more long-running processes, such as analyses and quality reports, these are interrupted by the update restart. 
Per our discussion, this PR reduces the update interval to once daily at 6am. 
This time point is after the docker registry replication time, hence, should have access to the newest docker images.

### DIFF
--- a/lib/systemd/bridgehead-update@.timer
+++ b/lib/systemd/bridgehead-update@.timer
@@ -2,7 +2,7 @@
 Description=Hourly Updates of Bridgehead (%i)
 
 [Timer]
-OnCalendar=*-*-* *:00:00
+OnCalendar=*-*-* 6:00:00
 
 [Install]
 WantedBy=basic.target


torbrenner (approved):

lib/systemd/bridgehead-update@.timer
```diff
@@ -2,7 +2,7 @@
 Description=Hourly Updates of Bridgehead (%i)
```
Threated, line 2:
```suggestion
Description=Daily Updates at 6am of Bridgehead (%i)
```

Threated (approved):

lib/systemd/bridgehead-update@.timer
```diff
@@ -1,8 +1,8 @@
 [Unit]
-Description=Hourly Updates of Bridgehead (%i)
+Description=Daily Updates at 6am of Bridgehead (%i)
 
 [Timer]
-OnCalendar=*-*-* *:00:00
+OnCalendar=*-*-* 6:00:00
```
lablans, line 6:
```suggestion
Persistent=true

```

If Bridgehead was down at 06:00, make sure the update is caught up on

lablans (changes requested):

lib/systemd/bridgehead-update@.timer
```diff
@@ -1,8 +1,8 @@
...
```
lablans, line 5:
```suggestion
OnCalendar=*-*-* 06:00:00
```

I've never seen it without two-digit numbers

lablans:
Are we not supposed to merge to develop branch first?

TKussel:
> Are we not supposed to merge to develop branch first?

GitHubs PR rebase makes this pretty ugly. But you are correct. As this change is not applied automatically, this ist currently not an issue but something to be mindful of in the future.

PR #202, opened 2024-08-15
Fixed environment variable passing for Directory sync

DavidCroftDKFZ:
There were problems with the passing of environment variables from bbmri.conf to the Directory synce container:

* The Directory password variable was misspellt.
* Some useful variables were missing.

Additionally, a delay was added before launching Directory sync, to give Blaze time to start up.

### DIFF
--- a/bbmri/modules/directory-sync-compose.yml
+++ b/bbmri/modules/directory-sync-compose.yml
@@ -4,12 +4,12 @@ services:
   directory_sync_service:
     image: "docker.verbis.dkfz.de/cache/samply/directory_sync_service"
     environment:
-      DS_DIRECTORY_URL: ${DS_DIRECTORY_URL}
+      DS_DIRECTORY_URL: ${DS_DIRECTORY_URL:-https://directory.bbmri-eric.eu}
       DS_DIRECTORY_USER_NAME: ${DS_DIRECTORY_USER_NAME}
       DS_DIRECTORY_USER_PASS: ${DS_DIRECTORY_USER_PASS}
-      DS_TIMER_CRON: ${DS_TIMER_CRON}
+      DS_TIMER_CRON: ${DS_TIMER_CRON:-0 22 * * *}
       # It is recommended to check the enabling of this flag with your local data protection group
-      DS_DIRECTORY_ALLOW_STAR_MODEL: ${DS_DIRECTORY_ALLOW_STAR_MODEL}
+      DS_DIRECTORY_ALLOW_STAR_MODEL: ${DS_DIRECTORY_ALLOW_STAR_MODEL:-true}
       DS_DIRECTORY_MOCK: ${DS_DIRECTORY_MOCK}
       DS_DIRECTORY_DEFAULT_COLLECTION_ID: ${DS_DIRECTORY_DEFAULT_COLLECTION_ID}
       DS_DIRECTORY_COUNTRY: ${DS_DIRECTORY_COUNTRY}
--- a/bbmri/modules/directory-sync-compose.yml
+++ b/bbmri/modules/directory-sync-compose.yml
@@ -6,7 +6,11 @@ services:
     environment:
       DS_DIRECTORY_URL: ${DS_DIRECTORY_URL}
       DS_DIRECTORY_USER_NAME: ${DS_DIRECTORY_USER_NAME}
-      DS_DIRECTORY_PASS_CODE: ${DS_DIRECTORY_PASS_CODE}
+      DS_DIRECTORY_USER_PASS: ${DS_DIRECTORY_USER_PASS}
       DS_TIMER_CRON: ${DS_TIMER_CRON}
       # It is recommended to check the enabling of this flag with your local data protection group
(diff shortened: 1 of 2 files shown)


patrickskowronekdkfz (approved):

Threated (changes requested):
If this service depends on blaze there should at least be a `depends_on:` besides where does it get the blaze url from? At some point we might also want to consider https://docs.docker.com/compose/startup-order/ so we dont have to sleep but thats a broader topic.

DavidCroftDKFZ:
The Blaze URL can be set via an environment variable, but the Bridgehead Blaze URL is the default for Directory sync, so does not need to be supplied

bbmri/modules/directory-sync-compose.yml
```diff
@@ -6,7 +6,11 @@ services:
...
       DS_DIRECTORY_USER_NAME: ${DS_DIRECTORY_USER_NAME}
-      DS_DIRECTORY_PASS_CODE: ${DS_DIRECTORY_PASS_CODE}
+      DS_DIRECTORY_USER_PASS: ${DS_DIRECTORY_USER_PASS}
       DS_TIMER_CRON: ${DS_TIMER_CRON}
       # It is recommended to check the enabling of this flag with your local data protection group
       DS_DIRECTORY_ALLOW_STAR_MODEL: ${DS_DIRECTORY_ALLOW_STAR_MODEL}
+      DS_DIRECTORY_MOCK: ${DS_DIRECTORY_MOCK}
+      DS_DIRECTORY_DEFAULT_COLLECTION_ID: ${DS_DIRECTORY_DEFAULT_COLLECTION_ID}
+      DS_DIRECTORY_COUNTRY: ${DS_DIRECTORY_COUNTRY}
+    command: sh -c "sleep 90 && java -jar directory_sync_service.jar" # Wait for Blaze before start
```
lablans, line 16: Blaze might not be up after 90 secs, or gone down in the meantime. All our apps should be resilient towards failures of other apps. Can you error-handle this in the service itself?

Also, since you obviously rely on Blaze, consider adding a dependency in Docker (not a guarantee but doesn't hurt)

bbmri/modules/directory-sync-compose.yml
```diff
@@ -6,7 +6,11 @@ services:
...
       DS_DIRECTORY_URL: ${DS_DIRECTORY_URL}
       DS_DIRECTORY_USER_NAME: ${DS_DIRECTORY_USER_NAME}
-      DS_DIRECTORY_PASS_CODE: ${DS_DIRECTORY_PASS_CODE}
+      DS_DIRECTORY_USER_PASS: ${DS_DIRECTORY_USER_PASS}
       DS_TIMER_CRON: ${DS_TIMER_CRON}
       # It is recommended to check the enabling of this flag with your local data protection group
       DS_DIRECTORY_ALLOW_STAR_MODEL: ${DS_DIRECTORY_ALLOW_STAR_MODEL}
+      DS_DIRECTORY_MOCK: ${DS_DIRECTORY_MOCK}
+      DS_DIRECTORY_DEFAULT_COLLECTION_ID: ${DS_DIRECTORY_DEFAULT_COLLECTION_ID}
+      DS_DIRECTORY_COUNTRY: ${DS_DIRECTORY_COUNTRY}
```
lablans, line 15: Where are these variables set in the Bridgehead?

bbmri/modules/directory-sync-compose.yml
```diff
@@ -6,7 +6,11 @@ services:
     environment:
       DS_DIRECTORY_URL: ${DS_DIRECTORY_URL}
       DS_DIRECTORY_USER_NAME: ${DS_DIRECTORY_USER_NAME}
-      DS_DIRECTORY_PASS_CODE: ${DS_DIRECTORY_PASS_CODE}
+      DS_DIRECTORY_USER_PASS: ${DS_DIRECTORY_USER_PASS}
       DS_TIMER_CRON: ${DS_TIMER_CRON}
       # It is recommended to check the enabling of this flag with your local data protection group
```
lablans, line 11: Noone will find this here and it's too vague to be helpful. Rather, describe what's happening in the readme or where the service is enabled (where IS it enabled?)

lablans (changes requested):

bbmri/modules/directory-sync-compose.yml
DavidCroftDKFZ: /etc/bridgehead/bbmri.conf

bbmri/modules/directory-sync-compose.yml
DavidCroftDKFZ: You say:

_Noone will find this here_

I am not sure I understand, could you give more details on the context of your statement?

The service is enabled in /etc/bridgehead/bbmri.conf.

Documented in: https://github.com/samply/bridgehead/?tab=readme-ov-file#directory-sync-tool

bbmri/modules/directory-sync-compose.yml
DavidCroftDKFZ: The service will retry if Blaze is not there.

I added a "depends_on" in my last git push.

lablans (approved):

PR #205, opened 2024-08-19
fix: specify host for id-management login

torbrenner:
Otherwise traefik will match the route with the one specified in datashield-compose.yml


patrickskowronekdkfz (approved):
Looks good to me. I'm wondering if the rule host works with IP addresses?

PR #222, opened 2024-09-03
Feature/pilot projects

PierreDelpy:


PierreDelpy:
This PR needs to be merged after develop is updated from main (it will then contain only 1 or 2 commits)

PR #230, opened 2024-09-13
refactor: put all modules in a common directory to remove redundancy

PierreDelpy:
This is a proposal to move all modules in a "common directory" so that we don't need code duplication.
The goal is to have a lightweight PR that does not change any volume names or similar. Just a drop in replacement to reduce code redundancy


Threated:
Great idea 🙏

patrickskowronekdkfz:
I like the idea. But this needs more work. I think the biggest hurdle will be configuration. We would need to make a good interface between the modules and the project configuration. For example having one module for gbn and bbmri connection (focus and beam) would be nice however, the configuration would be in the project. 

This would bring us closer to the idea of an template generator, but I'm not sure if we can put it in one pr.

PierreDelpy:
Closing this PR as it’s not feasible at this time. However, the idea of minimizing code redundancy is valuable, and we can keep it in our backlog for future consideration

PR #232, opened 2024-09-30
Feature/send branch to healthchecks

PierreDelpy:
added some refactoring to define the directories once and more importantly:
added monitoring for branches (lines 112-116)

### DIFF
--- a/lib/prerequisites.sh
+++ b/lib/prerequisites.sh
@@ -94,7 +94,7 @@ if [ "${PROJECT}" != "minimal" ]; then
   fi
 fi
 checkPrivKey() {
-  if [ -e "${CONFIG_DIR}"pki/"${SITE_ID}".priv.pem ]; then
+  if [ -e "${CONFIG_DIR}pki/${SITE_ID}.priv.pem" ]; then
     log INFO "Success - private key found."
   else
     log ERROR "Unable to find private key at ${CONFIG_DIR}pki/${SITE_ID}.priv.pem. To fix, please run\n  bridgehead enroll ${PROJECT}\nand follow the instructions."
--- a/lib/prerequisites.sh
+++ b/lib/prerequisites.sh
@@ -3,16 +3,16 @@
 source lib/functions.sh
 
 detectCompose
-configDirectory="/etc/bridgehead/"
-componentDirectory="/srv/docker/bridgehead/"
+CONFIG_DIR="/etc/bridgehead/"
+COMPONENT_DIR="/srv/docker/bridgehead/"
 
 if ! id "bridgehead" &>/dev/null; then
   log ERROR "User bridgehead does not exist. Please run bridgehead install $PROJECT"
   exit 1
 fi
 
-checkOwner ${configDirectory} bridgehead || exit 1
-checkOwner ${componentDirectory} bridgehead || exit 1
+checkOwner "${CONFIG_DIR}" bridgehead || exit 1
+checkOwner "${COMPONENT_DIR}" bridgehead || exit 1
 
 ## Check if user is a su
 log INFO "Checking if all prerequisites are met ..."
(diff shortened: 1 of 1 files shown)


Threated (approved):

lib/prerequisites.sh
```diff
@@ -32,39 +34,39 @@ fi
 log INFO "Checking configuration ..."
 
 ## Download submodule
-if [ ! -d "/etc/bridgehead/" ]; then
-  fail_and_report 1 "Please set up the config folder at /etc/bridgehead. Instruction are in the readme."
+if [ ! -d ${configDirectory} ]; then
```
TKussel, line 37:
```suggestion
if [ ! -d "${configDirectory}" ]; then
```

lib/prerequisites.sh
```diff
@@ -3,14 +3,16 @@
...
 
 if ! id "bridgehead" &>/dev/null; then
   log ERROR "User bridgehead does not exist. Please run bridgehead install $PROJECT"
   exit 1
 fi
 
-checkOwner /srv/docker/bridgehead bridgehead || exit 1
-checkOwner /etc/bridgehead bridgehead || exit 1
+checkOwner ${configDirectory} bridgehead || exit 1
+checkOwner ${componentDirectory} bridgehead || exit 1
```
TKussel, line 15:
```suggestion
checkOwner "${configDirectory}" bridgehead || exit 1
checkOwner "${componentDirectory}" bridgehead || exit 1
```

lib/prerequisites.sh
```diff
@@ -32,39 +34,39 @@ fi
...
   mkdir -p /etc/bridgehead/traefik-tls
 fi
 
-if [ ! -e "/etc/bridgehead/traefik-tls/fullchain.pem" ]; then
+if [ ! -e "${configDirectory}traefik-tls/fullchain.pem" ]; then
   openssl req -x509 -newkey rsa:4096 -nodes -keyout /etc/bridgehead/traefik-tls/privkey.pem -out /etc/bridgehead/traefik-tls/fullchain.pem -days 3650 -subj "/CN=$HOST"
 fi
 
-if [ -e /etc/bridgehead/vault.conf ]; then
+if [ -e $configDirectory}vault.conf ]; then
```
TKussel, line 61:
```suggestion
if [ -e "${configDirectory}"vault.conf ]; then
```

TKussel: Note the missing opening brace

lib/prerequisites.sh
```diff
@@ -32,39 +34,39 @@ fi
...
   if [ "$(stat -c "%a %U" /etc/bridgehead/vault.conf)" != "600 bridgehead" ]; then
     fail_and_report 1 "/etc/bridgehead/vault.conf has wrong owner/permissions. To correct this issue, run chmod 600 /etc/bridgehead/vault.conf && chown bridgehead /etc/bridgehead/vault.conf."
   fi
 fi
 
 log INFO "Checking network access ($BROKER_URL_FOR_PREREQ) ..."
 
-source /etc/bridgehead/${PROJECT}.conf
+source ${configDirectory}${PROJECT}.conf
```
TKussel, line 69:
```suggestion
source "${configDirectory}${PROJECT}".conf
```

lib/prerequisites.sh
```diff
@@ -92,10 +94,10 @@ if [ "${PROJECT}" != "minimal" ]; then
   fi
 fi
 checkPrivKey() {
-  if [ -e /etc/bridgehead/pki/${SITE_ID}.priv.pem ]; then
+  if [ -e ${configDirectory}pki/${SITE_ID}.priv.pem ]; then
```
TKussel, line 97:
```suggestion
  if [ -e "${configDirectory}"pki/"${SITE_ID}".priv.pem ]; then
```

lib/prerequisites.sh
```diff
@@ -107,6 +109,11 @@ else
   checkPrivKey || exit 1
 fi
 
+for dir in  ${configDirectory} ${componentDirectory}; do
```
TKussel, line 112:
```suggestion
for dir in  "${configDirectory}" "${componentDirectory}"; do
```

lib/prerequisites.sh
```diff
@@ -3,14 +3,16 @@
 source lib/functions.sh
 
 detectCompose
+configDirectory="/etc/bridgehead/"
+componentDirectory="/srv/docker/bridgehead/"
```
TKussel, line 7: Is the convention to use camel case variables for local variables? Otherwise, please convert to upper case

TKussel (changes requested):

patrickskowronekdkfz (changes requested):
I like the Idea of having the option to move the bridgehead in other directories. However, I would strongly advise to check if the path is valid. Also if you would change the path, this needs to be considered in the systemd files and the other scripts.

lib/prerequisites.sh
PierreDelpy: Thx for checking, added your suggestions

TKussel:
> I like the Idea of having the option to move the bridgehead in other directories. However, I would strongly advise to check if the path is valid. Also if you would change the path, this needs to be considered in the systemd files and the other scripts.

The config dir is explixitly checked. How does the "check ownership" function behave with non existing paths? Maybe this already handles the component dir.

I think this is a sensible change, even if systemd is not yet fully adapted (could be done by extracting the variables into a separate file and loding that as an env-file in the systemd service). Could that be done in a future PR?

TKussel (approved):

patrickskowronekdkfz (approved):

patrickskowronekdkfz:
> > I like the Idea of having the option to move the bridgehead in other directories. However, I would strongly advise to check if the path is valid. Also if you would change the path, this needs to be considered in the systemd files and the other scripts.
> 
> The config dir is explixitly checked. How does the "check ownership" function behave with non existing paths? Maybe this already handles the component dir.
> 
> I think this is a sensible change, even if systemd is not yet fully adapted (could be done by extracting the variables into a separate file and loding that as an env-file in the systemd service). Could that be done in a future PR?

Yes, I see it as a future PR.

lib/prerequisites.sh
```diff
@@ -92,10 +94,10 @@ if [ "${PROJECT}" != "minimal" ]; then
   fi
 fi
 checkPrivKey() {
-  if [ -e /etc/bridgehead/pki/${SITE_ID}.priv.pem ]; then
+  if [ -e "${CONFIG_DIR}"pki/"${SITE_ID}".priv.pem ]; then
```
Threated, line 97: This should be like so no?
```suggestion
  if [ -e "${CONFIG_DIR}pki/${SITE_ID}.priv.pem" ]; then
```

Threated (changes requested):

lib/prerequisites.sh
TKussel: It should expand to the same thing, but is the cleaner solution. I agree.

Threated (approved):

PR #233, opened 2024-10-01
Allow Usage of Centraxx Interface without login

torbrenner:
Before this change CentraXX was redirected to the central login servers then interacting with the id-management

### DIFF
--- a/ccp/modules/id-management-compose.yml
+++ b/ccp/modules/id-management-compose.yml
@@ -19,10 +19,18 @@ services:
       - traefik-forward-auth
     labels:
       - "traefik.enable=true"
+      # Router with Authentication
       - "traefik.http.routers.id-manager.rule=PathPrefix(`/id-manager`)"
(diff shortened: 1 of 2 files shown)


ccp/modules/id-management-compose.yml
```diff
@@ -57,7 +65,7 @@ services:
       - "/tmp/bridgehead/patientlist/:/docker-entrypoint-initdb.d/"
 
   traefik-forward-auth:
-    image: docker.verbis.dkfz.de/cache/oauth2-proxy/oauth2-proxy:v7.6.0
+    image: docker.verbis.dkfz.de/cache/oauth2-proxy/oauth2-proxy:latest
```
TKussel, line 68: Didn't Martin discuss explicit version tags at the department retreat? Is this applicable here?

torbrenner: The change was actually introduced by Martin. Reasoning is, that we currently don't have the versions in one place and should introduce that with another PR.

PR #239, opened 2024-10-15
feat: use project overarching vars

PierreDelpy:


vars
TKussel, line 1: `BEAM_TAG` is missing?

bridgehead
```diff
@@ -64,7 +64,7 @@ loadVars() {
 	fetchVarsFromVaultByFile /etc/bridgehead/$PROJECT.conf || fail_and_report 1 "Unable to fetchVarsFromVaultByFile"
 	setHostname
 	optimizeBlazeMemoryUsage
-	[ -e ./$PROJECT/vars ] && source ./$PROJECT/vars
+	[ -e ./$PROJECT/vars ] && source ./$PROJECT/vars ./vars
```
TKussel, line 67: Does that work for SecretSync as well @Threated ?

TKussel: @Threated tells me, that `loadVars` is executed before SecretSync :)

Threated:
So I dont know about the future use cases of the top level `vars` file but I think for now at least a name like `versions` or something  would be better

vars
TKussel: I'm not too happy about having different places where the tags are set, but we can tackle that in the future as it touches the `Environment` topic.

PierreDelpy:
> future use cases of the top level `vars` file but I think for now at least a name like `versions` or something would be better

you mean to rename the file or rename the variable names?

bridgehead
Threated, line 67: I think the source of ./vars should happen before the more specific project vars and regardless of project vars existing or not

Threated (changes requested):

Threated:
> > future use cases of the top level `vars` file but I think for now at least a name like `versions` or something would be better
> 
> you mean to rename the file or rename the variable names?

The file name

bridgehead
```diff
@@ -64,7 +64,7 @@ loadVars() {
 	fetchVarsFromVaultByFile /etc/bridgehead/$PROJECT.conf || fail_and_report 1 "Unable to fetchVarsFromVaultByFile"
 	setHostname
 	optimizeBlazeMemoryUsage
-	[ -e ./$PROJECT/vars ] && source ./$PROJECT/vars
+	[ -e ./$PROJECT/vars ] && source ./versions ./$PROJECT/vars
```
Threated, line 67: Sorry to bother you again I meant sourcing `./versions` regardless of `./$PROJECT/vars` existing. I think it always exits so it should not really matter but if we have the check that anyway

Threated (changes requested):
Also we should not need the `source versions` in each vars file as we source it before sourcing `$PROJECT/vars`

PR #241, opened 2024-10-22
feat: add focus tags in ccp and bbmri (#240)

enola-dkfz:


Threated (approved):

PierreDelpy:
Works fine for dktk, however the images size increased. I thought it should decrease?
grafik

Can someone please test develop branch in bbmri and then check if the search still works?

Threated:
No for dktk it will definitely increase because of the extra sql dependency for fhir2sql

enola-dkfz:
ERIC-test BH now runs on develop-bbmri, works well

PierreDelpy (approved):

PR #242, opened 2024-10-22
Update eric-compose.yml

enola-dkfz:
focus image gets dktk feature, not beam proxy


enola-dkfz:
not DKTK, BBMRI feature

Threated (approved):

PR #244, opened 2024-10-30
add and set FOCUS_ENDPOINT_TYPE

davidmscholz:
Add and set FOCUS_ENDPOINT_TYPE env variable depending on whether fhir2sql is active or not.


ccp/modules/fhir2sql-setup.sh
```diff
@@ -4,4 +4,5 @@ if [ "$ENABLE_FHIR2SQL" == true ]; then
   log INFO "Dashboard setup detected -- will start Dashboard backend and FHIR2SQL service."
   OVERRIDE+=" -f ./$PROJECT/modules/fhir2sql-compose.yml"
   DASHBOARD_DB_PASSWORD="$(generate_simple_password 'fhir2sql')"
+  export FOCUS_ENDPOINT_TYPE="blaze-and-sql"
```
Threated, line 7:
```suggestion
  FOCUS_ENDPOINT_TYPE="blaze-and-sql"
```
The export can be omitted because all the scripts are sourced while the `-a` flag is active which exports all bash vars

Threated (changes requested):

PR #246, opened 2024-11-05
set environment variables needed for the focus blaze-and-sql endpoint to work when fhir2sql is enabled

davidmscholz:


Threated (approved):
We might want to change the title to something more generic as this is not the only change from this PR

ccp/modules/id-management-compose.yml
```diff
@@ -101,5 +101,12 @@ services:
       forward_proxy:
         condition: service_healthy
 
+  ccp-patient-project-identificator:
+    image: samply/ccp-patient-project-identificator
+    container_name: bridgehead-ccp-patient-project-identificator
+    environment:
+      MAINZELLISTE_APIKEY: ${IDMANAGER_LOCAL_PATIENTLIST_APIKEY}
+      SITE_NAME: ${SITE_NAME}
+
```
TKussel, line 110: Is this supposed to be in this PR?

patrickskowronekdkfz: I would appreciate if this could make it into main with this PR. But we can move it to the next merge. It would add the PPI and it should not break anything.

TKussel: I would like it to be in a different PR for transparency reasons. However, we could merge them at the same-ish time.

ccp/modules/exporter-compose.yml
```diff
@@ -65,3 +65,8 @@ services:
       - "traefik.http.routers.reporter_ccp.tls=true"
       - "traefik.http.middlewares.reporter_ccp_strip.stripprefix.prefixes=/ccp-reporter"
       - "traefik.http.routers.reporter_ccp.middlewares=reporter_ccp_strip"
+
+  focus:
+    environment:
+      EXPORTER_URL: "http://exporter:8092"
+      AUTH_HEADER: "${EXPORTER_API_KEY}"
```
lablans, line 72: Is this set correctly? It says "header" in the key but the value only has an API key. [focus docs](https://github.com/samply/focus) state that there should be "ApiKey " in front

ccp/modules/id-management-compose.yml
lablans:
```suggestion
```
agreed, move to different PR

lablans (changes requested):

ccp/modules/exporter-compose.yml
enola-dkfz: @djuarezgf and I have agreed on the ApiKey type of the authentication some year ago. Is there a reason to have a different type?

TKussel: I looked into Focus' code: The Exporter uses a different header field (x-api-key) , so it is generally handled differently. This could be something to align in A) Explorter b) our other components using API-Keys. At the very least, this could and should be mentioned in Focus docs to be clear.

TKussel: @enola-dkfz thanks for jogging my memory. David chose the header recommended by OpenAPI, we chose the header recommended by the RFC and used throughout our components. Could you or @davidmscholz add a brief sentence to the focus documentation, that if the endpoint is exporter the semantics of the auth_header option changes?

enola-dkfz: Oh that's complicated, because when the target app is Exporter, the endpoint type is actually Blaze (or Blaze and SQL). As you said, Focus has outgrown its architecture.

TKussel: Thanks for adressing that in https://github.com/samply/focus/pull/185

TKussel (approved):

PR #247, opened 2024-11-05
Add exporter configuration to focus

djuarezgf:
Add exporter configuration in focus for  CCP in order to accept queries trough beam from central project manager (data science orchestrator)

PR #249, opened 2024-11-05
Add exporter configuration to focus

djuarezgf:
Add exporter configuration in focus for CCP in order to accept queries trough beam from central project manager (data science orchestrator)


PierreDelpy (approved):

Threated (changes requested):
Same complain from me as in #247

Threated (approved):

PR #255, opened 2024-12-12
Add support for SELinux systems

YtvwlD:
This is the other half of the changes we talked about last week: Since we're running the new VM with RHEL, we need some adjustments for SELinux. These are ignored on systems without it.

Adding [`Z`](https://docs.docker.com/reference/compose-file/services/#short-syntax-5) allows this container (and only this one) to access the bind mount. This seems to be only needed for files in `/srv`, because RHEL ships a special label for this folder.
This is also why `/srv/docker/bridgehead/bridgehead` can be run manually, but not from a systemd service.

We additionally had the slight problem that with Podman, `/var/run/docker.sock`, being a symlink to `/var/run/podman/podman.sock`, was owned by `root:root`. Changing this to `root:docker` worked, but I'm not quite sure where to put it in the existing scripts. (`docker` is the usual Docker cli.)

@Threated

### DIFF
--- a/minimal/modules/dnpm-node-compose.yml
+++ b/minimal/modules/dnpm-node-compose.yml
@@ -12,13 +12,13 @@ services:
       MYSQL_ROOT_HOST: "%"
       MYSQL_ROOT_PASSWORD: ${DNPM_MYSQL_ROOT_PASSWORD}
     volumes:
-      - /var/cache/bridgehead/dnpm/mysql:/var/lib/mysql
+      - /var/cache/bridgehead/dnpm/mysql:/var/lib/mysql:Z
 
   dnpm-authup:
     image: authup/authup:latest
     container_name: bridgehead-dnpm-authup
     volumes:
-      - /var/cache/bridgehead/dnpm/authup:/usr/src/app/writable
+      - /var/cache/bridgehead/dnpm/authup:/usr/src/app/writable:Z
     depends_on:
       dnpm-mysql:
         condition: service_healthy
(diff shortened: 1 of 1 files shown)


Threated:
Thanks a lot for the PR ❤️ 
Which volumes need the Z flag I saw you only added it to volumes that point to `/srv/docker/bridghead/*` so I guess `/etc/bridghead/*` is fine?
Also in other projects we have mounts in `/tmp/bridgehead/*` and `/var/cache/bridgehead/*` (I also changed the dnpm docker volumes to be fs mounts to that dir today as it was requested by Martin in the other PR) I assume that is problematic as well?
I wonder if we can get away with just executing a few commands as root when running `sudo ./bridgehead install <project>` to fix all the SEL specialties without needing to remember to add the z flag to each volume or not being able to use docker secrets 🤔

YtvwlD:
After taking a look at the contexts of the files, I'm not sure that we can solve this by just changing the files.
After cloning, everything is `unconfined_u:object_r:var_t:s0`.
But when the containers are running (with `:Z`), the files get different contexts:
`ccp/root.crt.pem` gets `system_u:object_r:container_file_t:s0:c252,c820`
`minimal/modules/dnpm-central-targets.json` gets `system_u:object_r:container_file_t:s0:c246,c734`
and these numbers change on container restart!

Someone with knowledge of selinux might be able to find a common label that makes the files available to all containers, but I can't. :) And I don't think a `Z` really hurts.

The `Z` is not required for files in `/etc`, I've removed it for `proxy.pem`.
I'm not sure why using Docker Secrets doesn't work here. Is this more than a normal mount?

I still need to take a look at `/var/cache`.

YtvwlD:
We seem to need the `Z` for `/var/cache` as well. The containers are not able to write to the folders otherwise.

TKussel:
@YtvwlD could you please rebase your branch based on the new `feat/selinux` branch? We are increasingly running into SELinux issues in other projects and would like to use your PR as a basis to extend it to all other projects as well.

PR #257, opened 2024-12-20
Use secret-sync for gitpassword

lablans:


lib/gitpassword.sh
```diff
@@ -1,41 +1,71 @@
...
+#echo "Called: $@" >> /tmp/credhelper
 
-BASE=$(baseDir)
-cd $BASE
+SECRETS_FILE=/var/cache/bridgehead/secrets/gitetcbridgehead
 
-source lib/functions.sh
+# Called from bridgehead, so shift 2 times
+shift
+shift
```
Threated, line 9: Not that it really matters but would not shifting before calling this make more sense?

lib/gitpassword.sh
```diff
@@ -1,41 +1,71 @@
...
+        -v $TMPFILE:/usr/local/cache \
+        -v $PRIVATEKEYFILENAME:/run/secrets/privkey.pem:ro \
+        -v /srv/docker/bridgehead/$PROJECT/root.crt.pem:/run/secrets/root.crt.pem:ro \
+        -v /etc/bridgehead/trusted-ca-certs:/conf/trusted-ca-certs:ro \
+        -e TLS_CA_CERTIFICATES_DIR=/conf/trusted-ca-certs \
+        -e NO_PROXY=localhost,127.0.0.1 \
+        -e ALL_PROXY=$HTTPS_PROXY_FULL_URL \
+        -e PROXY_ID=$PROXY_ID \
+        -e BROKER_URL=$BROKER_URL \
+        -e OIDC_PROVIDER=secret-sync-central.oidc-client-enrollment.$BROKER_ID \
```
Threated, line 45: This should be `gitlab_project_access_token_provider` in uppercase and set to the beam app id of the appropriate central service instead of setting an OIDC Provider

Threated (changes requested):

lib/gitpassword.sh
```diff
@@ -1,41 +1,71 @@
...
+        -e SECRET_DEFINITIONS=GitLabProjectAccessToken:GIT_CONFIG_REPO_TOKEN:bridgehead-configuration \
```
Threated, line 46: Also this is not how it is expected to be called I think it needs the full path to the repo which is of course a security concern as anyone with a valid beam cert could create a token for any other repo. Of course the disadvantage of auto generating the repo name from the beam task is that it might not match for some repos someone would have to check that but I guess its the better option

Threated (changes requested):

lib/gitpassword.sh
TKussel: Additionally: isn't `gitpassword.sh` also called from git as a credentials helper?

torbrenner: No, as we need in this script some environment variables setup in the bridgehead script (SITE_ID etc.), we decided that it would make more sense to provide this script as part of the bridgehead script. The git credential helper will be configured like this: `credential.helper="/srv/docker/bridgehead/bridgehead gitCredentials <project>"`
Because of this the 2 times `shift` is necessary to pass the inputs of git calling the credential helper to the gitpassword.sh script.

lib/gitpassword.sh
torbrenner: Yes, we discussed that this morning and decided to change the syntax of `SECRET_DEFINITIONS` to this. The idea is to derive the proper url of the sites configuration repository based on broker id and site id.

lib/gitpassword.sh
timakro: Yes, the name of the environment variable must be changed from `OIDC_PROVIDER` to `GITLAB_PROJECT_ACCESS_TOKEN_PROVIDER`. Martin configured it this morning such that the same central service generates OIDC tokens and GitLab project access tokens. So the beam id is probably correct although misleading.

timakro: Yes, the name of the environment variable must be changed from `OIDC_PROVIDER` to `GITLAB_PROJECT_ACCESS_TOKEN_PROVIDER`. Martin configured it this morning such that the same central service generates OIDC tokens and GitLab project access tokens. So the beam id is probably correct although misleading.

timakro (changes requested):

lib/gitpassword.sh
lablans: Fixed in b23615a

lib/gitpassword.sh
lablans: @Threated , is this resolved?

lablans:
Please note new commit 55fdeca0fcd557d3cf0890e0598cb3db2b6a7037, which will reconfigure the git repos to use the credential helper upon the next update.

lib/gitpassword.sh
Threated: No it should be: 
```suggestion
        -e GITLAB_PROJECT_ACCESS_TOKEN_PROVIDER=secret-sync-central.oidc-client-enrollment.$BROKER_ID \
```

timakro:
Only this one env var name change is missing. Otherwise looks good to me.

lib/functions.sh
```diff
@@ -116,7 +116,7 @@ assertVarsNotEmpty() {
 	MISSING_VARS=""
 
 	for VAR in $@; do
-	if [ -z "${!VAR}" ]; then
+		if [ -z "${VAR+x}" ] || [ -z "$VAR" ]; then
```
timakro, line 119: What is the intention of this change?

timakro:
I've moved the invocation of Secret Sync from `gitpassword.sh` to `update-bridgehead.sh` because any errors or log messages printed in `gitpassword.sh` are swallowed by git. The Secret Sync central component seems to be wrongly configured currently, so this is still a draft, but you an already review.

timakro (approved):

lib/functions.sh
timakro: This seems to break the original intention of the code so I changed it to how it was before.

timakro:
@lablans Can you review? I can't use the request review feature because this is your PR.

lib/update-bridgehead.sh
```diff
@@ -33,6 +33,35 @@ export SITE_ID
 checkOwner /srv/docker/bridgehead bridgehead || fail_and_report 1 "Update failed: Wrong permissions in /srv/docker/bridgehead"
 checkOwner /etc/bridgehead bridgehead || fail_and_report 1 "Update failed: Wrong permissions in /etc/bridgehead"
 
+# Use Secret Sync to validate the GitLab token in /var/cache/bridgehead/secrets/gitlab_token.
+# If it is missing or expired, Secret Sync will create a new token and write it to the file.
+# The git credential helper reads the token from the file during git pull.
+mkdir -p /var/cache/bridgehead/secrets
+touch /var/cache/bridgehead/secrets/gitlab_token # the file has to exist to be mounted correctly in the Docker container
+log "INFO" "Running Secret Sync for the GitLab token"
+docker pull docker.verbis.dkfz.de/cache/samply/secret-sync-local:latest # make sure we have the latest image
```
Threated, line 42: Do we really need this I thought `docker run image:latest` would check for the latest image every time?

timakro: It doesn't. I guess we could also do `docker run --pull=always`.

Threated: Oh really then I should update my secret sync invocation for oidc client creation as well because im pretty sure I dont have it there as well

timakro:
@Threated I made it a little bit nicer in https://github.com/samply/bridgehead/pull/257/commits/333591753e4bd871aab27dde1a3d059f1a0d96df

timakro:
Do you think this could take a long time to fail for projects other than ccp where there is no central secret sync running? That might be annoying.

Threated:
Very good point yeah. Its in the update job so its not that big of a deal I guess but we should probably fix that in secret sync itself and not in bash. I have an idea for it will dm you

timakro:
We talked about it and it should be fine as-is.

timakro:
We will do more testing because if this PR fails that could potentially stop updates to all bridgeheads. Jan has suggested some tests:
1. Test with the test bridgehead on `e260-serv-05`
2. Test in Mannheim
3. Test on BBMRI test
4. Try with a minimal config that does not have a beam proxy configured

Results:
1. Success. Token was created using secret sync and git credential helper was configured. Git pull works.
2. Not tested yet
3. Not tested yet
4. Not tested yet

timakro:
The tests where successful and we can merge. I'm not authorized to merge.

PR #258, opened 2025-01-10
feat: transFAIR

Threated:


modules/routine-connector-compose.yml
Threated, line 1: This being in a new top level folder is of course debatable and I am happy to move it to `dhki/modules` or something if the debate takes too long. Never the less we kind of need a concept for project Independent modules at some point

torbrenner (approved):

torbrenner:
tested on dkfz dhki bridgehead

PR #260, opened 2025-01-16
Update: Blaze to version 0.31

patrickskowronekdkfz:


Threated:
Didn't you say there was some breaking change/migration needed regarding ids or something and that's why we did not upgrade so far? Did that change?

patrickskowronekdkfz:
> Didn't you say there was some breaking change/migration needed regarding ids or something and that's why we did not upgrade so far? Did that change?

Yes, after the Update the indexes will change and we can't go back. However, I tested the version extensively and is stable. I would argue to make the upgrade

Threated (approved):

PR #264, opened 2025-01-27
New main release

Threated:


PierreDelpy (approved):

patrickskowronekdkfz (approved):
As Disscused, we should make a release. I don't remember the versioning for it, but I would suggest 2025.1

PR #265, opened 2025-01-27
feat: unify version handeling

### LINKED ISSUE #166: Add documentation for variable priorities
At present, bridgehead variables are established across various tiers:

Secrets
Vars
PROJECT.local.conf
PROJECT.conf
...
Maintaining clarity on the hierarchy of priorities for these variables would be beneficial in the event of overwriting, mitigating potential issues.

Threated:
I did not change any explicit versions yet. IMO that should happen in future PRs and we should focus on if this is the approach we want to go for.

Also added some docs for non bridgehead magicians.
So I guess closes #166

### DIFF
--- a/bridgehead
+++ b/bridgehead
@@ -53,17 +53,44 @@ case "$PROJECT" in
 		;;
 esac
 
+# Loads config variables and runs the projects setup script
 loadVars() {
-	# Load variables from /etc/bridgehead and /srv/docker/bridgehead
 	set -a
+	# Source the project specific config file
 	source /etc/bridgehead/$PROJECT.conf || fail_and_report 1 "/etc/bridgehead/$PROJECT.conf not found"
+	# Source the project specific local config file if present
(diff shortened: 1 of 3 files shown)


bridgehead
```diff
@@ -53,17 +53,44 @@ case "$PROJECT" in
...
+		"test")
+			source ./versions/test
+			;;
+		*)
+			report_error 7 "Environment \"$ENVIRONMENT\" is unknown. Assuming production. FIX THIS!"
+			source ./versions/prod
+			;;
+	esac
+	# Source the project specific config file
 	source /etc/bridgehead/$PROJECT.conf || fail_and_report 1 "/etc/bridgehead/$PROJECT.conf not found"
```
patrickskowronekdkfz, line 81: The project.conf should contain the var "ENVIROMENT". I think check the current branch is not a bad idea, but we should thunk about what is the default case.

patrickskowronekdkfz (approved):
We could think about renaming test -> develop, then we could remove the two files

PR #268, opened 2025-02-04
Main backport

Threated:
Nothing new here 3 merge commits and the other commits are the commits from #264 which are shown as a diff because I did a "rebase and merge" but that seems to trip git up. Now the [diff between develop and main](https://github.com/samply/bridgehead/compare/main...develop) still includes the commits from #264 which are both in develop and main. As we can't rebase develop because its protected we should just merge back main into develop before the diff grows larger. After that the above diff should only contain the 4 actually new commits I hope.


TKussel (approved):
So, no "rebase and merge" in develop/master branch anymore :D

Threated:
I think rebase and merge to merge into develop is fine. The problem just that we can't rebase develop on top of main because it's protected

PR #270, opened 2025-02-12
New Release to Main

torbrenner:
Changes:
- Removed restart: always from the docker compose files
- Updated Blaze to 0.31
- PPI now uses correct site id to query pseudonyms
- Added new Module for TransFAIR (currently only available in dhki)
- Added git password synchronization using Secret Sync
- Switched to new DNPM central node (@Threated please correct me if i'm wrong)


Threated:
The blaze update, PPI fix and the restart always thing are already in main. See #268 for why they still show up here.
> Switched to new DNPM central node (@Threated please correct me if i'm wrong)

Its not really a new central node as its not central but yeah its the new DNPM:dip node.

Threated (approved):

TKussel (approved):

PR #277, opened 2025-03-03
chore: unify blaze versioning and upgrade to 0.32

Threated:
Seems like there are no breaking changes https://github.com/samply/blaze/releases/tag/v0.32.0

### DIFF
--- a/versions/prod
+++ b/versions/prod
@@ -1,2 +1,3 @@
 FOCUS_TAG=main
 BEAM_TAG=main
+BLAZE_TAG=0.32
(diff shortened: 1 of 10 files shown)


versions/prod
```diff
@@ -1,2 +1,3 @@
 FOCUS_TAG=main
-BEAM_TAG=main
\ No newline at end of file
+BEAM_TAG=main
+BLAZE_TAG=0.32
```
patrickskowronekdkfz, line 3: This works fine, but maybe we should consider using latest?

patrickskowronekdkfz (approved):

versions/prod
Threated: I thought we wanted to have proper versioning for most of our components

torbrenner (approved):

PR #278, opened 2025-03-04
Send the origin of /etc/bridgehead repo to Secret Sync

timakro:
This is to anticipate an upcoming change in Secret Sync that will allow it to support multiple GitLab servers, e.g. verbis GitLab and BBMRI GitLab.

I moved things to a function and added the case statement. The only thing I changed in the code that was there before is this line:

```bash
-e SECRET_DEFINITIONS=GitLabProjectAccessToken:BRIDGEHEAD_CONFIG_REPO_TOKEN:$gitlab \
```

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -322,6 +322,57 @@ function sync_secrets() {
     set +a # Export variables in the regular way
 }
 
+function secret_sync_gitlab_token() {
+    # Map the origin of the git repository /etc/bridgehead to the prefix recognized by Secret Sync
+    local gitlab
+    case "$(git -C /etc/bridgehead remote get-url origin)" in
+        *git.verbis.dkfz.de*) gitlab=verbis;;
+        *gitlab.bbmri-eric.eu*) gitlab=bbmri;;
+        *)
+            log "WARN" "Not running Secret Sync because the git repository /etc/bridgehead has unknown origin"
+            return
+            ;;
+    esac
+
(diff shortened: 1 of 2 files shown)


lib/functions.sh
```diff
@@ -322,6 +322,57 @@ function sync_secrets() {
     set +a # Export variables in the regular way
 }
 
+function secret_sync_gitlab_token() {
+    # Map the origin of the git repository /etc/bridgehead to the prefix recognized by Secret Sync
+    local gitlab
+    case "$(git -C /etc/bridgehead remote get-url origin)" in
+        *git.verbis.dkfz.de*) gitlab=verbis;;
+        *gitlab.bbmri-eric.eu*) gitlab=bbmri;;
```
Threated, line 330: Do you think it might be a good idea if secret sync just took the full hostname as an argument instead of these special names. Then centrally we could look them up based on hostname?
Although looking at the secret sync code I think its fine as we can setup arbitrarily named gitlab providers and it might be a bit tricky do get the hostname of a url from bash.

timakro: Thought it was cleaner like this because otherwise how are we going to name the environment variables? `.` in the name of an environment variable feels like a compatibility nightmare.

Threated: Yeah I agree that could get ugly

TKussel (approved):

PR #281, opened 2025-03-10
Merge develop into main

timakro:
@Threated is moving the central Secret Sync from proxy ID `oidc-client-enrollment` to `central-secret-sync`. If secret sync fails for one or two nights that's no problem but within a week we should update the bridgeheads.


cce/docker-compose.yml
```diff
@@ -31,6 +32,10 @@ services:
       BEAM_PROXY_URL: http://beam-proxy:8081
       RETRY_COUNT: ${FOCUS_RETRY_COUNT}
       EPSILON: 0.28
+      QUERIES_TO_CACHE: '/queries_to_cache.conf'
+      ENDPOINT_TYPE: ${FOCUS_ENDPOINT_TYPE:-blaze}
+    volumes:
+      - /srv/docker/bridgehead/cce/queries_to_cache.conf:/queries_to_cache.conf
```
lablans, line 38: Make read-only?

Secondly, please know: If the file on the left doesn't exist, it will create an empty directory called queries_to_cache.conf.

lablans (changes requested):

torbrenner (approved):

cce/docker-compose.yml
lablans: With #282 now merged, this is resolved.

Threated:
This is kinda blocked on renaming the secret-sync instance from bbmri to `central-secret-sync` but it would be fine to merge as is as it *should* not break anything there

Threated (approved):

PR #285, opened 2025-03-17
chore!: update transfair config

Threated:
CC https://github.com/samply/transFAIR/pull/23


Threated:
Works with the samply/transfair:unify-ttp-args image of transfair now. Do we want to start versioning transfair from here on out? So should I set the transfair version in this PR to the 1.1 tag or something and when the transfair PR is merged we should tag that with 1.1? In this case it won't save us as the names of the vars in the config actually change now but it might be helpful in the future wdyt?

torbrenner (approved):

PR #286, opened 2025-03-18
Fix GitLab token syncing for BBMRI

timakro:
I don't think GitLab token syncing with Secret Sync ever worked for BBMRI because there is no `BROKER_URL` variable in BBMRI bridgeheads as it is called `ERIC_BROKER_URL` instead.

Is it fine if we merge this directly into main?


lib/functions.sh
```diff
@@ -334,6 +334,19 @@ function secret_sync_gitlab_token() {
             ;;
     esac
 
+    if [ "$PROJECT" == "bbmri" ]; then
+        # If the project is BBMRI, use the BBMRI-ERIC broker and not the GBN broker
+        proxy_id=$ERIC_PROXY_ID
+        broker_url=$ERIC_BROKER_URL
+        broker_id=$ERIC_BROKER_ID
+        root_crt_file="/srv/docker/bridgehead/bbmri/modules/${ERIC_ROOT_CERT}.root.crt.pem"
```
Threated, line 342: Does `ERIC_ROOT_CERT` really just contain the string `eric`? 😅

Threated (approved):

Threated:
Im fine with merging to main and back-porting to develop what do you think @torbrenner

lib/functions.sh
timakro: That or `eric.test`

Threated: Ahh

PR #292, opened 2025-04-14
feat: ssh tunnel

Threated:


TKussel (changes requested):
Can you please add documentation to the Readme.md? The usage and its requirements (e.g. non-passcode ssh key) are not self explanatory.

Threated:
Should we really add that documentation to the root level Readme? It does not contain any component specific docs yet. I feel like a markdown document in `ccp/modules` fits our current structure better (there are already markdown docs explaining some modules). Also I guess we could move this module to the top level modules folder as it is project independent.

TKussel:
True. The markdown documents in `*/modules` are a relict of the past and are discouraged, if I remember correctly. Moving it to `minimal/modules` is a good idea.

TKussel (approved):

Threated:
Its actually a top level modules folder (not in `minimal/`) Torben and I introduced with transfair.

PR #299, opened 2025-05-07
Use tmp directory for secret sync cache

timakro:
`/var/cache/bridgehead` did not exist on some bridgeheads, probably because they were installed before https://github.com/samply/bridgehead/blob/7365be3e7bf4d6ab7a9dcc4d5de40c8a13e5a8dc/lib/prepare-system.sh#L107 was added.

This PR uses a boot-scoped temporary directory to avoid such issues in the future.

The previous implementation kind of caused a mess on some servers because the Docker daemon created `/var/cache/bridgehead/secrets/gitlab_token` as a directory with owner `root:root`. Would be kind of difficult to clean up as most scripts run as `bridgehead` user.

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -348,7 +348,7 @@ function secret_sync_gitlab_token() {
     fi
 
     # Create a temporary directory for Secret Sync that is valid per boot
-    secret_sync_tempdir="/tmp/secret-sync.boot-$(cat /proc/sys/kernel/random/boot_id)"
+    secret_sync_tempdir="/tmp/bridgehead/secret-sync.boot-$(cat /proc/sys/kernel/random/boot_id)"
     mkdir -p $secret_sync_tempdir
 
     # Use Secret Sync to validate the GitLab token in $secret_sync_tempdir/cache.
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -347,18 +347,21 @@ function secret_sync_gitlab_token() {
         root_crt_file="/srv/docker/bridgehead/$PROJECT/root.crt.pem"
     fi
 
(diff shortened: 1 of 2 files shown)


lib/functions.sh
```diff
@@ -347,18 +347,21 @@ function secret_sync_gitlab_token() {
         root_crt_file="/srv/docker/bridgehead/$PROJECT/root.crt.pem"
     fi
 
-    # Use Secret Sync to validate the GitLab token in /var/cache/bridgehead/secrets/gitlab_token.
+    # Create a temporary directory for Secret Sync that is valid per boot
+    secret_sync_tempdir="/tmp/secret-sync.boot-$(cat /proc/sys/kernel/random/boot_id)"
```
lablans, line 351:
```suggestion
    secret_sync_tempdir="/tmp/bridgehead/secret-sync.boot-$(cat /proc/sys/kernel/random/boot_id)"
```

lib/gitlab-token-helper.sh
```diff
@@ -2,7 +2,7 @@
 
 [ "$1" = "get" ] || exit
 
-source /var/cache/bridgehead/secrets/gitlab_token
+source "/tmp/secret-sync.boot-$(cat /proc/sys/kernel/random/boot_id)/gitlab-token"
```
lablans, line 5:
```suggestion
source "/tmp/bridgehead/secret-sync.boot-$(cat /proc/sys/kernel/random/boot_id)/gitlab-token"
```

lablans (approved):

Threated:
I don't really see why we need the random boot id part?

timakro:
The plan was to start fresh even on distros that don't clear /tmp

PR #319, opened 2025-06-16
fix: adapt to transfair cli changes

Threated:
CC https://github.com/samply/transFAIR/pull/43


TKussel (approved):
Relates to https://github.com/samply/transFAIR/pull/43

PR #321, opened 2025-06-23
Add ghcr.io to URL list in README.md

pcvolkmer:


TKussel:
I would advice against replicating the DNPM:DIP in our registry. This feels like a mixing of responsibilities I am not comfortable with. Additinally, it's difficult to set a sensible replication period as the dev velocity is totally outside of our control.

TKussel (approved):

PR #327, opened 2025-07-04
Migrate CCP OIDC Configuration from Keycloak to Authentik

djuarezgf:
✅ Summary
This PR updates the OIDC configuration for the CCP project to use Authentik instead of Keycloak, aligning authentication with the current infrastructure standards.

📝 Details
- Replaces Keycloak with Authentik as the OIDC provider for all CCP components.

- The PSP component still uses Keycloak, as it was configured independently. This should also be updated in a future release.

- This change depends on the official release of the central-component **secret-sync** functionality, which is currently deployed manually from a test branch.

⚠️ Note
*Do not merge* this PR until the **secret-sync component is officially released**. This ensures compatibility and avoids configuration mismatches in production.

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -317,7 +317,6 @@ function sync_secrets() {
 
     mkdir -p /var/cache/bridgehead/secrets/ || fail_and_report 1 "Failed to create '/var/cache/bridgehead/secrets/'. Please run sudo './bridgehead install $PROJECT' again."
     touch /var/cache/bridgehead/secrets/oidc
-    echo $PRIVATEKEYFILENAME
     docker run --rm \
         -v /var/cache/bridgehead/secrets/oidc:/usr/local/cache \
         -v $PRIVATEKEYFILENAME:/run/secrets/privkey.pem:ro \


lib/functions.sh
```diff
@@ -301,19 +301,34 @@ function sync_secrets() {
...
+    else
+        proxy_id=$PROXY_ID
+        broker_url=$BROKER_URL
+        broker_id=$BROKER_ID
+        root_crt_file="/srv/docker/bridgehead/$PROJECT/root.crt.pem"
+    fi
+
     mkdir -p /var/cache/bridgehead/secrets/ || fail_and_report 1 "Failed to create '/var/cache/bridgehead/secrets/'. Please run sudo './bridgehead install $PROJECT' again."
     touch /var/cache/bridgehead/secrets/oidc
+    echo $PRIVATEKEYFILENAME
```
Threated, line 320:
```suggestion
```

lib/functions.sh
```diff
@@ -301,19 +301,34 @@ function sync_secrets() {
...
         -v /etc/bridgehead/trusted-ca-certs:/conf/trusted-ca-certs:ro \
         -e TLS_CA_CERTIFICATES_DIR=/conf/trusted-ca-certs \
         -e NO_PROXY=localhost,127.0.0.1 \
         -e ALL_PROXY=$HTTPS_PROXY_FULL_URL \
-        -e PROXY_ID=$PROXY_ID \
-        -e BROKER_URL=$BROKER_URL \
-        -e OIDC_PROVIDER=secret-sync-central.central-secret-sync.$BROKER_ID \
+        -e PROXY_ID=$proxy_id \
+        -e BROKER_URL=$broker_url \
+        -e OIDC_PROVIDER=secret-sync-central.test-secret-sync.$broker_id \
```
Threated, line 331: We should probably rename the proxy to oidc-secret-sync or something. Also not sure if we want to keep it on serv-05 but for now its fine I guess. But the rename is still a good idea before merging this I think

Threated (changes requested):

Threated (approved):

Issue #332, opened 2025-07-24
Replace hardcoded image: ...:develop references with version variables


djuarezgf:
## Problem

Currently, some services define their Docker image using a hardcoded `:develop` tag, like:

```yaml
image: samply/some-service:develop
````

This approach makes it harder to manage versions consistently across different environments.

## Proposal

We propose replacing hardcoded image tags with environment-specific variables. These variables are already defined in files under the `versions/` directory. For example:

### `versions/prod`

```bash
FOCUS_TAG=main
BEAM_TAG=main
BLAZE_TAG=0.32
POSTGRES_TAG=15.13-alpine
```

Each environment (`acceptance`, `prod`, `test`) has its own file that defines the necessary variables.

### Example change

Instead of:

```yaml
image: docker.verbis.dkfz.de/cache/samply/teiler-dashboard:develop
```

Use:

```yaml
image: docker.verbis.dkfz.de/cache/samply/teiler-dashboard:${TEILER_DASHBOARD_TAG}
```

## Tasks

* Identify all `image: ...:develop` references across the project.
* Replace them with the appropriate variable substitution (e.g., `${TEILER_DASHBOARD_TAG}`).
* Ensure that the correct environment-specific file from the `versions/` folder is loaded during deployment or CI.
* Test the setup to ensure the substitutions work correctly in all environments.

## Benefits

* Cleaner separation of code and environment-specific configuration
* Centralized version management
* Easier updates and rollbacks per environment
* Improved reproducibility and deployment stability

---

Feel free to share feedback or suggestions about this approach.

PR #335, opened 2025-07-28
Replace hardcoded image: ...:develop references with version variables

djuarezgf:
Implementation of issue https://github.com/samply/bridgehead/issues/332


djuarezgf:
Please, don't merge it yet, because I couldn't test it completely, as we are experiencing temporary problems with secret-sync.

Threated:
> **Note:** It only remains **beam-connect:develop**. @Threated : Should we also create a tag for it?

No

PR #337, opened 2025-08-11
feat(dnpm): change to new `api-gateway` image

Threated:
I removed the tag option for now as it does not have any other tags and I don't want to break deployments that use a custom tag


Threated:
I hope so too 🙃

PR #342, opened 2025-09-29
feat: add bridgehead check command

Threated:


lib/check-bridgehead.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+source lib/functions.sh
+
+log INFO "Running Bridgehead checks..."
+
+# Directory ownership
+log INFO "Checking directory ownership..."
+OWNERSHIP_OK=true
+if ! checkOwner /srv/docker/bridgehead bridgehead &> /dev/null; then
```
TKussel, line 9: Does this andthe following function calls need square brackets?

Threated: No

PR #382, opened 2026-05-07
add dnpm broker connector

Threated:
Depends on https://github.com/samply/beam-connect/pull/81


TKussel:
Thanks! This needs to be announced to the sites, so I converte this to a draft for now.

TKussel:
We should delete the committed central target, too @Threated ?

Threated:
Rebased the branch, removed the central targets and added the changes to the ccp dnpm files as well.

PR #388, opened 2026-05-29
latest blaze in acceptance and 1.8 in production 

enola-dkfz:
tested in test, it works
this sets blaze in acceptance and production to latest
however, this will not affect production until it is merged to main
it needs to be merged to develop first in any case


Threated:
I am a bit hesitant to change the prod tag to latest without in depth testing a real site first. I'm a not sure what might have changed in blaze between these version

enola-dkfz:
I tested in ERIC-test BH, but if you have access to a Production BH please do test it. The Blaze change is from the end of last year, we had a CQL function Diagnosis which checked for diagnosis both in Conditions and Specimens. There was a coalesce, but that wasn't the problem, according to Alex, the very definition of a function with more than one parameter is what doesn't work in newer Blaze versions. 

(We used to check in Specimen as well because Würzburg didn't have Conditions at all, only ICD10 codes in specimens. It is a workaround we have meant to remove for 2 years at least.)

Threated:
I am not concerned about that. It's more about potential compatibility problems in the data model of blaze itself

enola-dkfz:
IIRC it was only ever used in BBMRI. Other projects' CQLs should have been compatible with the new Blaze all along.

Threated:
Yeah that is not what I am concerned about

enola-dkfz:
> Yeah that is not what I am concerned about

Do you have access to a PROD BH using CQL to test with latest blaze?

enola-dkfz:
will change it to v1.8 for now and periodically update

Threated (approved):

PR #390, opened 2026-06-17
new blaze version and project stuff

enola-dkfz:


Threated (approved):

itcc/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   blaze:
-    image: docker.verbis.dkfz.de/cache/samply/blaze:${BLAZE_TAG}
+    image: docker.verbis.dkfz.de/cache/samply/blaze:latest
```
TKussel, line 5: Shouldn't that use the blaze tag?

TKussel (changes requested):

itcc/docker-compose.yml
TKussel: @Martin1088 ?

Threated: AFAIK they needed a new blaze feature or something but now that #388 is merged I think it can be changed back to the variable.

enola-dkfz: we need to merge today, so that BHs update tomorrow so that we are here if something doesn't work

TKussel (approved):

PR #391, opened 2026-06-23
added CQL_FLAVOUR

enola-dkfz:
variable that tells Focus which CQL "flavour" to use when translating AST to CQL
when not set (in most circumstances) Focus uses the project name from the task metadata


Threated (commented):
I think I would rather call that bridgehead config var something with focus in the name

enola-dkfz:
Are there other BH components translating into CQL?

Threated:
No but it's also focus retry count for example. I think it just makes it more obvious for people writing/reading the configs

enola-dkfz:
fine by me

Threated (approved):

TKussel (approved):

